PR #176, opened 2020-06-21
Terraform automated set-up of composer environment

aroelo:
This set-up creates 3 workspaces:
- A composer 'dev' workspace
- A composer 'prod' workspace
- A workspace with shared resources

The workspaces are created by using a prefix ('ao-' for now) and a name ('dev', 'prod', 'shared' resp.). The shared workspace is created from the 'shared' subdirectory, while the composer workspaces are created from the terraform root directory. 

I have moved all the dockerfiles in a separate 'dockerfiles' directory, because it became a bit chaotic to have them all in the root directory.

Some issues/TODO's:
- There are some values used in Terraform that can't be retrieved from variables. To keep track of these values and in some cases use them from the python application I stored them in the `TerraformConfig` class. Perhaps there is a better way to do this than storing them in a class.
- Decide whether we use terraform to create an external project and set-up resources in there, or assume that we already have a project in which we want set-up the resources.
- Decide how to get the pypi packages that need to be installed inside the composer environment.
- When terraform cloud is >= version 0.13.* we can use 'count' in the composer module definition, instead of using 'count' in each resource definition that is inside the module. 
- Not sure if the `terraform` directory should be inside `academic_observatory`, maybe it can be moved up one directory to `academic-observatory`

Resolves https://github.com/The-Academic-Observatory/academic-observatory/issues/112
Resolves https://github.com/The-Academic-Observatory/academic-observatory/issues/111
Resolves https://github.com/The-Academic-Observatory/academic-observatory/issues/105
Resolves https://github.com/The-Academic-Observatory/academic-observatory/issues/59
Resolves https://github.com/The-Academic-Observatory/academic-observatory/issues/54


aroelo:
Deprecated, closing this pull request.

PR #199, opened 2020-07-06
POC: Terraform turn on/off vm from DAG

aroelo:
Some DAG logic (might expand more later):

There are separate dags for turning the vm on and off. 
The dag to turn the vm on will be run based on a schedule.
The dag to turn the vm off will be triggered externally (at the end of each telescope dag, see the dummy telescope for an example).

The dags use the same TerraformWorkspace class (which is now in the telescopes directory, for a lack of better location atm), since they use mostly the same methods.

The dag to turn the vm on will push a xcom unit with a datetime object (called 'start_time_vm'), which can be pulled in the dag to turn the vm off.

The dag to turn the vm off should be triggered each time a telescope is finished. When it is triggered it will get all dag ids and their corresponding states. If a dag was started after 'start_time_vm', it will check if that dag state is successful. If all dag states are successful, the vm will be turned off.

I added functionality for using a target addrs, but it seems that this doesn't work when the resource has not been created yet.

Terraform set-up to make this work:
```
# set-up terraform credentials
terraform login
# init
terraform init
# create workspaces
terraform workspace new dev
terraform workspace new staging
terraform workspace new prod
```

Terraform Cloud:
* Add variables to each workspace
  * google-credentials (sensitive)
  * project ('workflows-dev')
  * region ('us-east1')
  * zone ('us-east1-b')
  * vm-status ('off')

* Set each workspace to auto-apply

```
# select & apply each workspace
terraform workspace select dev
terraform workspace apply
terraform workspace select staging
terraform workspace apply
terraform workspace select prod
terraform workspace apply
```


aroelo:
Deprecated, see https://github.com/The-Academic-Observatory/observatory-platform/pull/231 for implementation of more recent version.

PR #224, opened 2020-07-21
Ignore docs and notebooks directories for language classification

jdddog:
Only include .py files for language classification

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
 tests/fixtures/** filter=lfs diff=lfs merge=lfs -text
 * linguist-vendored
-*.ipynb linguist-vendored=false
+*.py linguist-vendored=false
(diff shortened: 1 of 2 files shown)


jdddog:
> Just reading about github linguist now.
> Does this mean that everything is excluded (vendored), except for any *.py files?
> So the project would then probably be 100% python?

Yeah, Jupyter Notebooks store a lot of binary content which takes up a lot of space (marking most of the project as a Jupyter Notebook, when it isn't), which is why I wanted to ignore the .ipynb files.

After thinking about it, it might be better to ignore the docs and notebooks directories instead, then if any other languages are used they will be picked up (the changes in the latest push).

rhosking:
> After thinking about it, it might be better to ignore the docs and notebooks directories instead, then if any other languages are used they will be picked up (the changes in the latest push).

agreed

Issue #222, opened 2020-07-21
Install instructions Observatory Platform Mac OS


aroelo:
As mentioned in PR https://github.com/The-Academic-Observatory/observatory-platform/pull/206, we need to check the current instructions to install Observatory Platform on Mac OS and new instructions to install the dependencies 'packer' and 'terraform'

> Yeah that would be good.
> 
> There are some instructions for installing the Observatory Platform on Mac OS: https://observatory-platform.readthedocs.io/en/latest/installation.html#system-dependencies-macos.
> 
> The instructions for installing Packer and Terraform are now in a tutorial on Read the Docs: https://observatory-platform.readthedocs.io/en/latest/deploy_terraform.html#linux. We should also add the installation instructions for Mac OS as a section here.
> 
> Would someone with Mac OS mind testing these instructions and adding the ones for Packer and Terraform into the deploy Terraform tutorial?
> 

See https://github.com/The-Academic-Observatory/observatory-platform/pull/206#discussion_r457705958

**Installing brew**
Should be updated, the output from the current command is:  
```
284060a@M-A0042882:~/repos/observatory-platform$ /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
Warning: The Ruby Homebrew installer is now deprecated and has been rewritten in
Bash. Please migrate to the following command:
  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```

Update with:
`/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"`

**Installing python3.7**
Just tried `brew install python3.7` according to the current documentation. This is giving an error:
```
284060a@M-A0042882:~/repos/observatory-platform$ brew install python3.7
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (homebrew/core).
==> New Formulae
arb                                                                gateway-go                                                         scw@1
==> Updated Formulae
ammonite-repl            aws-cdk                  daemontools              dwarfutils               ghc@8.8                  mariadb-connector-odbc   nodenv                   vert.x
anime-downloader         bde                      dbmate                   folly                    gitlab-runner            mitmproxy                polyml                   wxmaxima
avfs                     cedille                  dnsmasq                  gh                       mage                     netdata                  scw                      z.lua

Error: No available formula with the name "python3.7" 
[...]

PR #227, opened 2020-07-22
Update installation instructions in README.md for mac OS environment.

aroelo:
Didn't realise in the previous PR (https://github.com/The-Academic-Observatory/observatory-platform/pull/223) that there are also separate install instructions in the readme. This PR updates the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,27 +61,27 @@ Install Docker Engine:
 ### System dependencies: MacOS
 Install [Homebrew](https://brew.sh/) with the following command:
 ```bash
-/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 ```
 
 Install Python 3.7 with brew:
 ```bash
-brew install python3.7
+brew install sashkab/python/python@3.7
 ```
 
-Install pip:
+Add Python 3.7 to path:
 ```bash
-curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
-python3.7 get-pip.py --user
+echo 'export PATH="/usr/local/opt/python@3.7/bin:$PATH"' >> ~/.bash_profile
 ```
 
 Install virtualenv 20 or greater:
 ```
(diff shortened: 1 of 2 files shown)


aroelo:
> Thanks Aniek, this looks good.
> 
> Is there any particular reason for using this particular Python package?
> 
> ```
> brew install sashkab/python/python@3.7
> ```

From #222 

> **Installing python3.7**
> Just tried `brew install python3.7` according to the current documentation. This is giving an error:
> 
> ```
> 284060a@M-A0042882:~/repos/observatory-platform$ brew install python3.7
> Updating Homebrew...
> ==> Auto-updated Homebrew!
> Updated 1 tap (homebrew/core).
> ==> New Formulae
> arb                                                                gateway-go                                                         scw@1
> ==> Updated Formulae
> ammonite-repl            aws-cdk                  daemontools              dwarfutils               ghc@8.8                  mariadb-connector-odbc   nodenv                   vert.x
> anime-downloader         bde                      dbmate                   folly                    gitlab-runner            mitmproxy                polyml                   wxmaxima
> avfs                     cedille                  dnsmasq                  gh                       mage                     netdata                  scw                      z.lua
> 
> Error: No available formula with the name "python3.7" 
> ==> Searching for a previously deleted formula (in the last month)...
> Error: No previously deleted formula found.
> ==> Searching for similarly named formulae...
> This similarly named formula was found:
> python@3.7
> To install it, run:
>   brew install python@3.7
> ==> Searching taps...
> ==> Searching taps on GitHub...
> Error: No formulae found in taps.
> ```
> 
> I don't think it's possible to specify a version like that with brew, see also https://apple.stackexchange.com/questions/237430/how-to-install-specific-version-of-python-on-os-x
> 
> This might be an alternative: `brew install sashkab/python/python@3.7` (see https://github.com/sashkab/homebrew-python)
> 
> Have to add it to the path:
> `echo 'export PATH="/usr/local/opt/python@3.7/bin:$PATH"' >> /Users/284060a/.bash_profile`

Just looked into it a bit further, when I try `brew install python@3.7` instead of `brew install python3.7` it does work. Reading from the thread earlier I thought using the sashkab package was the easiest way to install a specific python version, but if `brew install python@3.7` works this is better.

Issue #121, opened 2020-06-01
Setting up separate terraform google cloud project 


aroelo:
Partially following https://cloud.google.com/community/tutorials/managing-gcp-projects-with-terraform

Create project: 
`terraform-admin`, project id: `terraform-admin-279004`

Linked billing account

Create service account:
```
gcloud iam service-accounts create terraform \
>   --display-name "Terraform admin account"
```

Create & Download keys json
Stored locally:
`/Users/aniekroelofs/Downloads/terraform-admin-279004-2e72b7feddc4.json`

Grant the service account permission to view the Admin Project and manage Cloud Storage:
```
export TF_ADMIN=terraform-admin-279004

gcloud projects add-iam-policy-binding ${TF_ADMIN} \
  --member serviceAccount:terraform@${TF_ADMIN}.iam.gserviceaccount.com \
  --role roles/viewer

gcloud projects add-iam-policy-binding ${TF_ADMIN} \
  --member serviceAccount:terraform@${TF_ADMIN}.iam.gserviceaccount.com \
  --role roles/storage.admin
```

TODO
Any actions that Terraform performs require that the API be enabled to do so. Possible list:
```
gcloud services enable cloudresourcemanager.googleapis.com
gcloud services enable cloudbilling.googleapis.com
gcloud services enable iam.googleapis.com
gcloud services enable compute.googleapis.com
gcloud services enable serviceusage.googleapis.com
gcloud services enable composer.googleapis.com
gcloud services enable bigquery.googleapis.com
gcloud services enable bigquerystorage.googleapis.com
```

TODO
Grant the service account permission to create projects and assign billing accounts:
```
export TF_VAR_org_id=670938844927

gcloud organizations add-iam-policy-binding ${TF_VAR_org_id} \
  --member serviceAccount:terraform@${TF_ADMIN}.iam.gserviceaccount.com \
  --role roles/resourcemanager.projectCreator

gcloud organizations add-iam-policy-binding ${TF_VAR_org_id} \
  --member serviceAccount:terraform@${TF_ADMIN}.iam.gserviceaccount.com \
  --role roles/billing.user
```

`ERROR: (gcloud.organizations.add-iam-policy-binding) User [aniek.roelofs@observatory.academy] does not have permission to access organization [670938844927:getIamPolicy] (or it may not exist): The caller does not have permission`

TODO
Create bucket & enable object versioning:
```
[...]

PR #231, opened 2020-07-24
Terraform on cloud, turning VM on/off, adding APIs

aroelo:
I have to add proper documentation for functions etc, but the code is working and thought it would be a good moment to review the content.

Added features:
* **Added `data.google_compute_subnetwork.ao_subnetwork`**

The subnetwork in `module.airflow_main_vm.google_compute_address.vm_private_ip` forced replacement every time `terraform apply` was executed.
Consequently, `module.airflow_main_vm.google_compute_instance.vm_instance` also had to be replaced.
Explicitly referencing to the data object `data.google_compute_subnetwork.ao_subnetwork` prevents this. The data object has to depend on the ao_network, so it is forced to be 'read' by Terraform upon apply.

* **Added APIs, see also https://github.com/The-Academic-Observatory/observatory-platform/issues/121#issuecomment-660794573**

All required APIs are now enabled inside the terraform configuration, with the exception of the compute engine API.

This API has to be enabled already for packer to build the image. It can't be destroyed upon `terraform destroy`, because the existing image prevents this.
We could either:
1. Leave out this API completely from Terraform configuration. Assuming that it has been enabled already.
2. Include this API in terraform configuration and set disable_on_destroy=True
3. Include this API in terraform configuration and find out how to delete compute image

Currently I've implemented option 2. It might be nice to be delete the compute image & destroy the API if the API itself costs money.

* **Store terraform workspace on Terraform Cloud**

The full name of a workspace in terraform cloud consists of a prefix and suffix. The suffix is based on the project id.

Replaced variable 'credentials_file' with multi-line 'credentials' variable. Terraform Cloud has to store the values inside the cloud workspace and this can't be a file.

It is necessary to explicitly enter the terraform directory to execute all the terraform commands, because of files that otherwise can't be found:
```
  on main.tf line 527, in data "template_file" "airflow_main_vm_startup":
 527:   template = file("terraform/startup-main.tpl")
 and
   on main.tf line 564, in data "template_file" "airflow_worker_vm_startup":
 564:   template = file("./terraform/startup-worker.tpl")
```

* **Create/destroy airflow worker VM based on variable**

* **Integrate VM creation into airflow**

Added separate dags to create/destroy worker vm.
The dag to create the worker vm can be executed based on a schedule interval.
The dag to destroy the worker vm will be triggered externally (at the end of each telescope dag, see the dummy telescope for an example).

The dags use the same TerraformWorkspace class (which is now in the telescopes directory, not sure if this is the best location), since they use mostly the same methods.

The dag to create the vm will push a xcom unit with a datetime object (called 'start_time_vm'), which can be pulled in the dag to turn destroy the vm.

The dag to destroy the vm is triggered each time a telescope is finished. When it is triggered it will get all dag ids and their corresponding states. If a dag was started after 'start_time_vm', it will check if that dag state is successful. If all dag states are successful, the vm will be destroyed.

@jdddog when testing out which APIs were required I got to the following required roles on the service account:
Cloud SQL Admin
Compute Admin
Compute Image User
Compute Network Admin
Create Service Accounts
Delete Service Accounts
Service Account Key Admin
Service Account User
Project IAM Admin
Secret Manager Admin
Service Usage Admin
Storage Admin
Storage Transfer Admin

Which are more than there are listed in the docs. Are the roles in the documentation additional to the 'editor' role? I'm planning to try out the set-up with the roles that you specified in the documentation, but didn't get to this yet and perhaps you will get to reading this review first.


aroelo:
> When terraform init is run, you can pass in a customisable backend configuration. This means the name of the terraform organisation can be changed. It needs these files etc.
> 
> backend.hcl:
> ```
> hostname = "app.terraform.io"
> organization = "jamie-test"
> ```
> 
> Changes to the terraform setting in main.tf (it might be better to call the workspace prefix observatory now):
> ```
> terraform {
>   backend "remote" {
>     workspaces {
>       prefix = "observatory-"
>     }
>   }
> }
> ```
> 
> Then you run. So you can change the backend file if you want to use another organisation.
> ```
> terraform init -backend-config=backend.hcl
> ```

Nice, I think this is a great idea. The organisation name is now hardcoded, if we use click like you suggested we can add the organisation as an argument at this step.

> Also, since we are creating the program to upload the variables to the workspace, it might be more consistent to use the same type of yaml configuration file that we use for running the observatory locally. It would also mean that we could yaml to mark variables as secrets rather than using comments, e.g. add secret:True to each variable that you want to be a secret or secrets: and then all of the secret variables like what you have. I'm not sure what is best. It would also mean that you don't have to copy the service account key contents into the file, as you could use Python to read the contents.

This sounds good. The only reason I used the .tfvars file format is if we would like to be able to use terraform with a local state as well, so you could still use the same variables file. I agree that it is a lot nicer to use variables to mark secrets instead of comments.

> The last thing we need to do is to prevent certain resources from being destroyed when running the system in production mode, e.g. we don't want the database or the storage buckets on the production system to get wiped out by mistake. One way to do this is to set prevent_destroy=True, for instance:
> 
> ```
> lifecycle {
>     prevent_destroy = true
> }
> ```
> The resources to add it to include these types of resources:
> 
> google_sql_database_instance
> google_sql_database
> google_sql_user
> google_storage_bucket
> Perhaps these could be set from a variable so that we can still destroy these instances in the development and testing environments but not in production?
>
> Unfortunately this isn't fool proof, because apparently if you remove the resource from the file or rename it the flag isn't respected and the resource get's deleted anyway! See hashicorp/terraform#17599.
> 
> According to the commentor here, the only way to prevent this is to not give the service account that Terraform uses permission to destroy certain types of resources. So perhaps for the production instance we need to work out a narrower set of permissions to give the service account that Terraform uses, this would be for managing: Cloud Storage and Cloud SQL so that it can't destroy those resources even if it tries.

Yeah better safe than sorry with something like that :-). Using the workspace suffix might be a good idea to use for this (instead of a separate variable). The suffix is now obtained from the project id. Something like this:
```
lifecycle {
    prevent_destroy = terraform.workspace == "production_project_id" ? true : false
  }
```

aroelo:
@jdddog I made some changes regarding the backend for 'terraform init' and adding a click command to create the terraform workspace. Maybe you can have a look at these changes when you have some time and I will keep working on the destroying of resources part.

aroelo:
> * The command line interface could be simplified a little bit by making it nested one more level:
> ```
> @cli.group()
> def terraform():
>     pass
> 
> @terraform.command()
> def create_workspaces(config_path, terraform_variables_path, organization, prefix, terraform_credentials_file, debug):
> ```

Ah, nice, didn't know that was possible!

> * config-path: perhaps this should be an argument rather than an option, because there will 3 config files (one for each of dev, staging and prod), so it will prevent mistakes by making the user provide the path to the config file.

Okay, that makes sense. Would you prefer an argument over a required option? I thought using an option gives some more clarity on what file you need, but an argument might be nice, because it's shorter. 

> * You can also change the "type" parameter for config-file to click.Path(exists=True, file_okay=True, dir_okay=False) and then the checks to see if the file exists don't need to be done further down in the code.

Ah okay, I just made it kind of similar to the config file of the `observatory platform` command. If you set it to click.Path(exists=True, file_okay=True, dir_okay=False) I thought that the code will exit when the file doesn't exists and you won't be able to generate a default config file.

> * The "organization" parameter could be added as a parameter to terraform-config.yaml and removed as a command line argument. This also means that the user has to fill it in before running the command as well.

This might be related to the last question:

> What is the terraform-variables-path file used for?

The variables path is used to check which variables are used inside the terraform configuration and thus should also be defined in the config file. 

This has some benefits of:
* Being able to create a schema with required variables
* Generating a default config file from the schema
* If google_application_credentials is in the schema, it will check if the file exists (similar to observatory platform)
* Copying the descriptions from the variables file and uploading these to the cloud workspace

I thought I would keep the config file limited to only terraform cloud variables, so it directly correlates to the variables file, but I could add the organization and probably hardcode it in the schema. 

> * Rather than prefix + project id, workspaces created could be one of the below options. The dev, staging and prod suffixes can be supplied from the "environment" variable in the config file:
>    * prefix + "dev"
>    * prefix + "staging"
>    * prefix + "prod" 

The way I saw it the 'environment' variable is used inside airflow to determine which files are processed, e.g. when the environment is 'dev' the smaller test files will be used when developing, however, you might want to set the environment to 'prod' inside the 'dev' google cloud project to test your DAG with the 'real' files? I might have the wrong idea about this.

Would you develop the pipeline with smaller test files and when this is successful use the 'real' file in staging environment only? Or is there another way to do this that I'm missing.

aroelo:
> The last thing we need to do is to prevent certain resources from being destroyed when running the system in production mode, e.g. we don't want the database or the storage buckets on the production system to get wiped out by mistake. One way to do this is to set prevent_destroy=True, for instance:
> 
> ```
> lifecycle {
>     prevent_destroy = true
> }
> ```
> The resources to add it to include these types of resources:
> 
> google_sql_database_instance
> google_sql_database
> google_sql_user
> google_storage_bucket
> Perhaps these could be set from a variable so that we can still destroy these instances in the development and testing environments but not in production?
> 
> Unfortunately this isn't fool proof, because apparently if you remove the resource from the file or rename it the flag isn't respected and the resource get's deleted anyway! See hashicorp/terraform#17599.
> 
> According to the commentor here, the only way to prevent this is to not give the service account that Terraform uses permission to destroy certain types of resources. So perhaps for the production instance we need to work out a narrower set of permissions to give the service account that Terraform uses, this would be for managing: Cloud Storage and Cloud SQL so that it can't destroy those resources even if it tries.
> 
**Reducing permissions as backup plan to prevent destruction of resources**
Required permissions to create database instance and sql user:
`cloudsql.users.create` and `cloudsql.instances.create`

The role with fewest permissions that include these is the sql admin role. However, this role also includes permissions to delete resources. I've made a custom role from the standard sql editor role with the addition of adding `cloudsql.users.create` and `cloudsql.instances.create`.

image

I created a similar role for the storage buckets by taking the storage admin role and removing the `storage.buckets.delete` and `storage.objects.delete` permissions.
image

The complete list of assigned roles is:
image


**Assigning lifecycle to prevent_destroy=true**
Apparently you can't use a variable inside the 'prevent_destroy' declaration. See https://github.com/hashicorp/terraform/issues/3116. The issue was closed quite a while ago, but I couldn't find any updates/fixes for this. 

For now I've implemented the idea of this workaround mentioned here: https://github.com/hashicorp/terraform/issues/3116#issuecomment-292038781.

The only catch is that it would be possible to delete the "random_id" resources that is used for naming of the database & buckets. This could be done when both the workaround "random_id" and the "random_id" that is used for naming are commented out or removed from the code.

Terraform won't delete any databases/buckets then, however because the names of these resources are lost, when re-applying Terraform it will try to recreate the databases/buckets. It won't be able to do this, because it can't delete the current versions, so Terraform will be stuck in this case.

aroelo:
@jdddog changes since last review:
* Updated prefix from ao- to observatory-
* Suffix is now obtained from 'environment' variable instead of 'project_id'. 
* Changed config path from click option to click argument
* Removed 'organization' click option, added to config file instead. The 'organization' variable in the config file is the only variable that isn't uploaded to terraform cloud and has a separate category: api.
* Created use of nested config file, see deploy_terraform.md for example. The config file can only be nested 1 level deep. It is assumed that the main variable name doesn't have any other properties such as 'category' or 'value', only nested subvariables.
The main and subvariables in the config file are created by splitting the terraform variables on a '-'.
* Removed random ids from database&bucket names
* Added `update_workspace` command that will overwrite existing variables using the config file
* Initially simplified command line interface by making it nested one more level like you suggested, however when I added the `update_workspace` command it made more sense to change it back and have two commands `create_workspace` and `update_workspace` 
* Added workaround for terraform to prevent destroy of database & bucket resources in case of a 'prod' environment
* I looked into setting custom roles/permissions to prevent destroying of database & bucket resources, as mentioned above and as we discussed briefly via chat. This isn't really a code change, but if you think this is the way we would like to go I will update the documentation accordingly.

Please let me know if some things are unclear :-).

observatory_platform/telescopes/terraform.py
```diff
@@ -0,0 +1,144 @@
...
+        prefix = Variable.get(AirflowVar.terraform_prefix.get())
+        environment = Variable.get(AirflowVar.environment.get())
+        workspace = prefix + environment
+
+        # Get workspace ID
+        workspace_id = terraform_api.workspace_id(organization, workspace)
+
+        # Push messages
+        ti: TaskInstance = kwargs['ti']
+        ti.xcom_push(TerraformTelescope.XCOM_WORKSPACE_ID, workspace_id)
```
jdddog, line 76: I wonder if it would be cleaner to wrap this functionality up as a function and get the workspace id, rather than using xcoms to get the workspace id within each task?

jdddog:
> I looked into setting custom roles/permissions to prevent destroying of database & bucket resources, as mentioned above and as we discussed briefly via chat. This isn't really a code change, but if you think this is the way we would like to go I will update the documentation accordingly.

Yeah this sounds like the way to go.

jdddog:
I had to make a few changes (see the last commit) so that the terraform_organization and terraform_prefix variables were created as secrets.

When I run the create_vm DAG I get the following error in the get_workspace_id? I have set the airflow_connection/terraform value to a Terraform API user token.
```
[2020-08-19 00:30:06,009] {terraform_utils.py:70} INFO - Response status: 401
[2020-08-19 00:30:06,010] {terraform_utils.py:71} INFO - Unsuccessful retrieving workspace id for workspace 'observatory-staging', response: {"errors":[{"status":"401","title":"unauthorized"}]}
```

Do you think that you could deploy and test the create and destroy VM DAGs and see if they still run for you?

aroelo:
> I had to make a few changes (see the last commit) so that the terraform_organization and terraform_prefix variables were created as secrets.
> 
> When I run the create_vm DAG I get the following error in the get_workspace_id task? I have set the airflow_connection/terraform value to a Terraform API user token.
> 
> ```
> [2020-08-19 00:30:06,009] {terraform_utils.py:70} INFO - Response status: 401
> [2020-08-19 00:30:06,010] {terraform_utils.py:71} INFO - Unsuccessful retrieving workspace id for workspace 'observatory-staging', response: {"errors":[{"status":"401","title":"unauthorized"}]}
> ```
> 
> Would you be able to deploy and test the create and destroy VM DAGs and see if they still run for you?

I'm recreating some terraform resources and will look into it now :-).

observatory_platform/telescopes/terraform.py
aroelo: Thanks for the idea, my reasoning was that I would try to limit the number of requests made to the terraform API, but you're right that it would be cleaner!

aroelo: See https://github.com/The-Academic-Observatory/observatory-platform/pull/231/commits/df09eec07da0812782fd780a4c45e5a0c0a976a3

aroelo:
> * What do you think about consolidating the functionality of TerraformConfig and ObservatoryConfig and perhaps the syntax of the files?

Good question. Just to be clear, do you mean we'll still have two separate config files? But the code to read/validate/create default files for both is consolidated under one class?

I think it would be probably be a lot clearer and more practical to consolidate them. The main reason I created TerraformConfig separate to ObservatoryConfig was because of the different schema. Parts of the schema are hardcoded before the ObservatoryConfig class is initialised, so we'll have to make this more flexible depending on which variables are required. 

Also, having the TerraformConfig separate I could then directly use the 'variables.tf' file to create a schema that can be used for validation, this seemed like a nice solution to me. However, with the now nested values in the config file the schema validation got a bit more complicated and ugly. 

Perhaps it might be better to drop the use of 'variables.tf' and create a separate class similar to AirflowConn and AirflowVar where we hardcode the required variables. This would make it easier to consolidate TerraformConfig/ObservatoryConfig as well.

jdddog:
> When I run the create_vm DAG I get the following error in the get_workspace_id task? I have set the airflow_connection/terraform value to a Terraform API user token.

This was my mistake, I didn't specify the connection string properly!

These are the list of things that we talked about on Slack that would be good to look at:
* We should set the time interval for these DAGs
  * The create VM DAG should turn the VM on, on the same weekly / monthly cycle as the DAGs that run on the worker VM. It can be at the same time (i.e. it doesn't have to start early), if that makes it simpler to maintain because the DAG run tasks will be queued when the worker VM is off.
  * Perhaps the destroy VM DAG should run every 10 minutes?
* Add an additional task for vm_create and vm_destroy which waits until the apply plan has finished and is set to success or failed if the apply plan succeeded or failed.
* The ability to explicitly specify a list of DAG ids that the vm_destroy DAG has to wait to be finish (or something to handle this, we can discuss more).
* What do you think about adding a task that sends a Slack message that says whether the apply plan failed or succeeded (for the vm create and destroy tasks)? Like in the Airflow tutorials.

> What do you think about consolidating the functionality of TerraformConfig and ObservatoryConfig and perhaps the syntax of the files?
> Good question. Just to be clear, do you mean we'll still have two separate config files? But the code to read/validate/create default files for both is consolidated under one class?

Yeah there would still be two different config files.

There are a lot of common variables between running the observatory locally and running it with Terraform, so it would be good if the config files were as similar as possible and used the same class to load them.

> I think it would be probably be a lot clearer and more practical to consolidate them. The main reason I created TerraformConfig separate to ObservatoryConfig was because of the different schema. Parts of the schema are hardcoded before the ObservatoryConfig class is initialised, so we'll have to make this more flexible depending on which variables are required.

Yeah, I'll have a think about this and see if I can give some suggestions.

aroelo:
>   * The create VM DAG should turn the VM on, on the same weekly / monthly cycle as the DAGs that run on the worker VM. It can be at the same time (i.e. it doesn't have to start early), if that makes it simpler to maintain because the DAG run tasks will be queued when the worker VM is off.

What if the DAGs that run on the worker VM have a different cycle? How do we pick which DAGs the VM is in sync with?
So as long as the worker VM is off the DAGs won't fail, but just wait to be run?

>   * Perhaps the destroy VM DAG should run every 10 minutes?

This seems like a good idea, I've added this. 

> * Add an additional task for vm_create and vm_destroy which waits until the apply plan has finished and is set to success or failed if the apply plan succeeded or failed.

Terraform Cloud itself already queues plans if earlier plans are not finished yet. I can imagine adding an additional task could give us some more control if we incorporate it with other DAGs/tasks or just give an idea of the status. Do you want to set the task to 'failed' if the apply failed and get some insights this way? Or should it be success as long as the apply finished? 

> * The ability to explicitly specify a list of DAG ids that the vm_destroy DAG has to wait to be finish (or something to handle this, we can discuss more).

I've added an implementation of this in my latest commit (will link once it's pushed) maybe you can have a look at this? How will we maintain the list of DAG ids. Should this be done automatically based on whether a DAG has a queue specified to 'remote_queue', or do we manually add DAG ids?
 
> * What do you think about adding a task that sends a Slack message that says whether the apply plan failed or succeeded (for the vm create and destroy tasks)? Like in the Airflow tutorials.

Yeah I think that's great, maybe we could do that for other Telescopes in production as well. It would be nice to get a heads-up when something fails.

jdddog:
For the config files, perhaps something like this would work, I'm probably not thinking of all of the implications though:

The config file used to run the local environment:
```yaml
backend: local
project_id: academic-observatory-staging
region: us-west1
zone: us-west1-c
data_location: us
environment: staging
google_application_credentials: /path/to/creds/credentials.json

database:
  backup_start_time: "23:00"
  tier: "db-custom-2-7680"

airflow_main:
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_worker:
  create: false
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_ui:
  user_email: user@email.com
  user_password: password

airflow_secret:
  fernet_key: fernetkey
  postgres_password: password
  redis_password: password

airflow_connection:
  mag_releases_table: mysql://myname:mypassword@myhost.com
  mag_snapshots_container: mysql://myname:mypassword@myhost.com
  crossref: mysql://myname:mypassword@myhost.com
```

The config file used to deploy the system with terraform:
```yaml
backend:
  terraform: # Use the organization value with the Terraform API and upload the organization and workspaces_prefix as variables that will be created as airflow secrets
    organization: jamie-test
    workspaces_prefix: observatory-

project_id: academic-observatory-staging
region: us-west1
zone: us-west1-c
data_location: us
environment: staging
google_application_credentials:
  value: /path/to/creds/credentials.json
  file: true
  sensitive: true

database:
  backup_start_time: "23:00"
  tier: "db-custom-2-7680"

airflow_main:
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_worker:
  create: false
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_ui:
  user_email: user@email.com
  user_password:
    value: password
    sensitive: true

airflow_secret:
  fernet_key:
    value: fernetkey
    sensitive: true
  postgres_password:
    value: password
    sensitive: true
  redis_password:
    value: password
    sensitive: true

airflow_connection:
  mag_releases_table:
    value: mysql://myname:mypassword@myhost.com
    sensitive: true
  mag_snapshots_container:
    value: mysql://myname:mypassword@myhost.com
    sensitive: true
  crossref:
    value: mysql://myname:mypassword@myhost.com
    sensitive: true
  terraform:
    value: terraform: mysql://terraform:mypassword@myhost.com
    sensitive: true
```

The main difference is that you would have to specify a "backend", either local (the local dev environment) or terraform. 
* If the backend is terraform then you also need to specify the organization and workspaces_prefix. The organization is used by the create-workspace and update-workspace commands to interact with the Terraform API. The organization and workspaces_prefix should also be uploaded as variables so that they can be loaded as secrets into Google Secrets Manager to be used by Airflow DAGs as variables.
* This also means that we don't need to set the category of each variable to terraform when using the config file for terraform, making the config file a bit cleaner. If the backend is set to terraform then all values can be uploaded as variables to Terraform (except for perhaps the backend which is a special case).

Marking variables as sensitive:
* This can be the same as how it is currently.
* If the local backend is specified, then sensitive can be ignored (but still parse).

You should be able to specify the values for variables, by setting the value key or just directly setting the value for the variable:
```yaml
airflow_secret:
  fernet_key: # setting the value key
    value: fernetkey
  postgres_password: password # directly setting the variable
```
* It will make the config files cleaner when you don't need to specify additional properties for variables (e.g. sensitive).

I think that we should validate that 'fernet_key', 'postgres_password' and 'redis_password' have been set for 'airflow_secret', since they are all required to run the syste,.

I don't think we need to validate that certain connections under 'airflow_connection' have been specified, just that the connection URI strings have been specified correctly (there must be a library that can already do this validation).

jdddog:
For the config files, perhaps something like this would work, I'm probably not thinking of all of the implications though:

The config file used to run the local environment:
```yaml
backend: local
project_id: academic-observatory-staging
region: us-west1
zone: us-west1-c
data_location: us
environment: staging
google_application_credentials: /path/to/creds/credentials.json

database:
  backup_start_time: "23:00"
  tier: "db-custom-2-7680"

airflow_main:
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_worker:
  create: false
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_ui:
  user_email: user@email.com
  user_password: password

airflow_secret:
  fernet_key: fernetkey
  postgres_password: password
  redis_password: password

airflow_connection:
  mag_releases_table: mysql://myname:mypassword@myhost.com
  mag_snapshots_container: mysql://myname:mypassword@myhost.com
  crossref: mysql://myname:mypassword@myhost.com
```

The config file used to deploy the system with terraform:
```yaml
backend:
  terraform: # Use the organization value with the Terraform API and upload the organization and workspaces_prefix as variables that will be created as airflow secrets
    organization: org-name
    workspaces_prefix: observatory-

project_id: academic-observatory-staging
region: us-west1
zone: us-west1-c
data_location: us
environment: staging
google_application_credentials:
  value: /path/to/creds/credentials.json
  file: true
  sensitive: true

database:
  backup_start_time: "23:00"
  tier: "db-custom-2-7680"

airflow_main:
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_worker:
  create: false
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_ui:
  user_email: user@email.com
  user_password:
    value: password
    sensitive: true

airflow_secret:
  fernet_key:
    value: fernetkey
    sensitive: true
  postgres_password:
    value: password
    sensitive: true
  redis_password:
    value: password
    sensitive: true

airflow_connection:
  mag_releases_table:
    value: mysql://myname:mypassword@myhost.com
    sensitive: true
  mag_snapshots_container:
    value: mysql://myname:mypassword@myhost.com
    sensitive: true
  crossref:
    value: mysql://myname:mypassword@myhost.com
    sensitive: true
  terraform:
    value: terraform: mysql://terraform:mypassword@myhost.com
    sensitive: true
```

The main difference is that you would have to specify a "backend", either local (the local dev environment) or terraform. 
* If the backend is terraform then you also need to specify the organization and workspaces_prefix. The organization is used by the create-workspace and update-workspace commands to interact with the Terraform API. The organization and workspaces_prefix should also be uploaded as variables so that they can be loaded as secrets into Google Secrets Manager to be used by Airflow DAGs as variables.
* This also means that we don't need to set the category of each variable to terraform when using the config file for terraform, making the config file a bit cleaner. If the backend is set to terraform then all values can be uploaded as variables to Terraform (except for perhaps the backend which is a special case).

Marking variables as sensitive:
* This can be the same as how it is currently.
* If the local backend is specified, then sensitive can be ignored (but still parse).

You should be able to specify the values for variables, by setting the value key or just directly setting the value for the variable:
```yaml
airflow_secret:
  fernet_key: # setting the value key
    value: fernetkey
  postgres_password: password # directly setting the variable
```
* It will make the config files cleaner when you don't need to specify additional properties for variables (e.g. sensitive).

I think that we should validate that 'fernet_key', 'postgres_password' and 'redis_password' have been set for 'airflow_secret', since they are all required to run the syste,.

I don't think we need to validate that certain connections under 'airflow_connection' have been specified, just that the connection URI strings have been specified correctly (there must be a library that can already do this validation).

aroelo:
Thanks for tackling this together and coming up with some great ideas!

> The main difference is that you would have to specify a "backend", either local (the local dev environment) or terraform.
> 
> * If the backend is terraform then you also need to specify the organization and workspaces_prefix. The organization is used by the create-workspace and update-workspace commands to interact with the Terraform API. The organization and workspaces_prefix should also be uploaded as variables so that they can be loaded as secrets into Google Secrets Manager to be used by Airflow DAGs as variables.

The 'organization' and 'workspaces_prefix' we need as airflow variables for turning the VM on/off. These should also be available inside the local observatory platform environment. So maybe we should add these to that config file as well?

> * This also means that we don't need to set the category of each variable to terraform when using the config file for terraform, making the config file a bit cleaner. If the backend is set to terraform then all values can be uploaded as variables to Terraform (except for perhaps the backend which is a special case).

The 'terraform' category was more meant to distinguish terraform 'terraform' and terraform 'env' variables (https://www.terraform.io/docs/cloud/workspaces/variables.html#terraform-variables). I don't think we'll be using any 'env' variables soon, so 'terraform' could probably just be the default and it can be left out of the config :-).

> Marking variables as sensitive:
> 
> * This can be the same as how it is currently.
> * If the local backend is specified, then sensitive can be ignored (but still parse).
> 
> You should be able to specify the values for variables, by setting the value key or just directly setting the value for the variable:
> 
> ```yaml
> airflow_secret:
>   fernet_key: # setting the value key
>     value: fernetkey
>   postgres_password: password # directly setting the variable
> ```
> 
> * It will make the config files cleaner when you don't need to specify additional properties for variables (e.g. sensitive).

Yeah, nice, that looks a lot better. 

> I think that we should validate that 'fernet_key', 'postgres_password' and 'redis_password' have been set for 'airflow_secret', since they are all required to run the system.

That makes sense, will add that to the schema!

> I don't think we need to validate that certain connections under 'airflow_connection' have been specified, just that the connection URI strings have been specified correctly, because you may or may not want to run the DAGs that use them (there must be a library that can already do this validation). I wonder if you can iterate through variables in Terraform? Then you could dynamically create the connection secrets depending on what the user specified.

Do you mean using the 'variables.tf' file for iterating through variables? 

In the config file for the local environment. How do we distinguish between airflow variables, airflow connections and other variables? This is related to how the environment variables 'AIRFLOW_VAR_* and 'AIRFLOW_CONN_*' are automatically created. 

From the local config, these we don't need for the local environment, right? Or would you like to keep them in so that the config files for terraform/local are the same as much as possible? But we wouldn't need them as airflow variables I think.
```yaml
database:
  backup_start_time: "23:00"
  tier: "db-custom-2-7680"

airflow_main:
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_worker:
  create: false
  disk_size: 20
  disk_type: pd-standard
  machine_type: n2-standard-2

airflow_ui:
  user_email: user@email.com
  user_password: password
```

Perhaps we should just put these under an 'airflow_variables' header.
```yaml
backend: local
project_id: academic-observatory-staging
region: us-west1
zone: us-west1-c
data_location: us
environment: staging
google_application_credentials: /path/to/creds/credentials.json
```

Simplified local config file.
```yaml
backend: local

airflow_variables:
  project_id: academic-observatory-staging
  region: us-west1
  zone: us-west1-c
  data_location: us
  environment: staging
  google_application_credentials: /path/to/creds/credentials.json
  fernet_key: fernetkey
  postgres_password: password
  redis_password: password
  terraform_organization: org-name
  terraform_workspaces_prefix: observatory-

airflow_connection:
  mag_releases_table: mysql://myname:mypassword@myhost.com
  mag_snapshots_container: mysql://myname:mypassword@myhost.com
  crossref: mysql://myname:mypassword@myhost.com
```

jdddog:
> Thanks for tackling this together and coming up with some great ideas!
> 
> > The main difference is that you would have to specify a "backend", either local (the local dev environment) or terraform.
> > 
> > * If the backend is terraform then you also need to specify the organization and workspaces_prefix. The organization is used by the create-workspace and update-workspace commands to interact with the Terraform API. The organization and workspaces_prefix should also be uploaded as variables so that they can be loaded as secrets into Google Secrets Manager to be used by Airflow DAGs as variables.
> 
> The 'organization' and 'workspaces_prefix' we need as airflow variables for turning the VM on/off. These should also be available inside the local observatory platform environment. So maybe we should add these to that config file as well?

In the local environment the "remote worker" Docker container is always on so it doesn't need turning on and off.


> > I don't think we need to validate that certain connections under 'airflow_connection' have been specified, just that the connection URI strings have been specified correctly, because you may or may not want to run the DAGs that use them (there must be a library that can already do this validation). I wonder if you can iterate through variables in Terraform? Then you could dynamically create the connection secrets depending on what the user specified.
> 
> Do you mean using the 'variables.tf' file for iterating through variables?


If the connections could be represented in the Terraform variables.tf as a collection (e.g. a list or map) then in main.tf, one could loop over the connections to create each secret (if you can loop over things), rather than having secret module instances (like below) that are hard coded. I'm not sure if that is possible though. There are collection types: https://www.terraform.io/docs/configuration/types.html#collection-types

```
module "mag_releases_table_connection_secret" {
  source = "./secret"
  secret_id = "airflow-connections-mag_releases_table"
  secret_data = var.mag_releases_table_connection
  service_account_email = google_service_account.ao_service_account.email
}
```


> From the local config, these we don't need for the local environment, right? Or would you like to keep them in so that the config files for terraform/local are the same as much as possible? But we wouldn't need them as airflow variables I think.
> 
> ```yaml
> database:
>   backup_start_time: "23:00"
>   tier: "db-custom-2-7680"
> 
> airflow_main:
>   disk_size: 20
>   disk_type: pd-standard
>   machine_type: n2-standard-2
> 
> airflow_worker:
>   create: false
>   disk_size: 20
>   disk_type: pd-standard
>   machine_type: n2-standard-2
> 
> airflow_ui:
>   user_email: user@email.com
>   user_password: password
> ```
> 

Ah you're right, that was my mistake, they would not be needed locally.

> ```yaml
> backend: local
> 
> airflow_variables:
>   project_id: academic-observatory-staging
>   region: us-west1
>   zone: us-west1-c
>   data_location: us
>   environment: staging
>   google_application_credentials: /path/to/creds/credentials.json
>   fernet_key: fernetkey
>   postgres_password: password
>   redis_password: password
>   terraform_organization: org-name
>   terraform_workspaces_prefix: observatory-
> 
> airflow_connection:
>   mag_releases_table: mysql://myname:mypassword@myhost.com
>   mag_snapshots_container: mysql://myname:mypassword@myhost.com
>   crossref: mysql://myname:mypassword@myhost.com
> ```

That looks good, the only thing is that some of those don't end up being airflow variables. project_id, data_location, environment (and then terraform_workspaces_prefix and terraform_organization are when terraform is used). The others become environment variables or are just used by Terraform.

aroelo:
@jdddog, my apologies that the last commit has become a bit massive. However, since most parts are connected it was hard to do it in smaller separate bits. Here's a bit of an overview of changes since your last review.
Config changes:  
- Consolidated the `TerraformConfig` class into `ObservatoryConfig` (I actually wrote a new config class for this, but it's still named ObservatoryConfig) 
- Left out `__eq__(self, other`) and `__ne__(self, other)` methods, since I wasn't sure what they are used for exactly.
- Merged `gen_config_terraform` into `gen_config_interface`
- Validates connection uri and gives an error when an airflow Connection object can't be created with the given uri. Should it give a warning when there is no password/extra set? E.g. accidentally using `mysql://crossref-token@myhost.com` instead of `mysql://:crossref-token@myhost.com` doesn't give an error, but it might be useful to point out that there is no password set.
- Use of default values for examples, but no actual default values set in the schema, so that a value has to be filled in for each variable. For the local config the default values are hardcoded inside config_utils, for the terraform config they are hardcoded inside the `variables.tf` file. The default values defined in variables.tf will be overwritten by the values stored in Terraform cloud, so they aren't actually used in a set-up. 
- Use '!sensitive' yaml tags for sensitive values in terraform config. I found that adding this as a tag made it easier to validate/create the schema, because it links more easily to the `variables.tf` file. If we would occasionally add fields for 'sensitive', the variable would become a 'dict' type instead of 'string', while a 'string' is expected. 
- If a value is an existing path in terraform config, the content is automatically uploaded to Terraform cloud as a string.
- Added `stream_process` function which streams the stdout/stderr from a subprocess, instead of printing the output once the subprocess is finished. This makes it easier to spot when a process is stuck at a step it shouldn't be stuck out (e.g. with the large host-uid number bug)

Terraform changes:
- Use `templatefile()` function to create template file instead of `template_file` resource. Terraform recommends using this newer function instead. This was also necessary to be able to pass different types of vars to the template file (such as map()).
- Create google project services with a for_each loop, reducing code clutter.
- Create airflow secrets and airflow connection secrets using for_each loop.
- Airflow variables are now loaded as environment variables by including them in `startup-main.tpl` and `startup-worker.tpl`. They are dynamically exported as environment variables, so that it is easier to add more variables in the future.
- Some variables in variables.tf have become 'object' or 'map' variables, this is necessary to use 'for_each' on the variable and also makes it easier to validate/create a corresponding config file.
- Use random id in database name when not in production environment.
- Added default value for each variable.

Other:
- Added custom airflow Variable class which catches the exception for trying to get (non-existing) google secrets with insufficient permissions. It's now added to `generic_utils.py`, not sure if this is the most appropriate location. The class could easily be imported with `from observatory_platform.utils.generic_utils import AirflowVariable as Variable` in existing DAGs/telescopes. I haven't replaced this yet.
- Add SlackWebhookHook to send a slack message when the terraform run status is not 'applied' or 'planned_and_finished'. Have to think about a solution on sharing the slack webhook token.
- For a list of given DAG ids check whether these DAGs have all finished successfully (since the VM has been turned on).
- Add function/task that checks the run status of a Terraform run and returns when the run is finished (successful or not).

Final local config file:
```yaml
backend: "local"
fernet_key: "yb3V_7rn6WlTV-dNN5LShPEOlc0khtDg4lb5-E_8JFk="
google_application_credentials: "/path/to/credentials/file/json"
airflow_connections:
  crossref: "mysql://:crossref-token@"
  mag_releases_table: "mysql://azure-storage-account-name:url-encoded-sas-token@"
  mag_snapshots_container: "mysql://azure-storage-account-name:url-encoded-sas-token@"
  terraform: "mysql://:terraform-token@"
  slack: "https://https%3A%2F%2Fhooks.slack.com%2Fservices?webhook_token=/T00000000/B00000000/XXXXXXXXXXXXXXXXXXXXXXXX"
airflow_variables:
  environment: "dev"
  project_id: "your-project-id"
  data_location: #"us"
  download_bucket_name: "your-bucket-name"
  transform_bucket_name: "your-bucket-name"
  terraform_organization: "your-organization"
  terraform_prefix: "observatory-"
```

Final terraform config file:
```yaml
backend:
  terraform:
    organization: "your-organization"
    workspace_prefix: "observatory-"

project_id: "your-project-id"
!sensitive google_application_credentials: "/path/to/service/account/key.json"
environment: "dev"
region: "us-west1"
zone: "us-west1-c"
data_location: "us"

database:
  tier: "db-custom-4-15360"
  backup_start_time: "23:00"

airflow_main:
  machine_type: "n2-standard-2"
  disk_size: "20"
  disk_type: "pd-ssd"

airflow_worker:
  machine_type: "n1-standard-2"
  disk_size: "20"
  disk_type: "pd-standard"

airflow_worker_create: "False"

!sensitive airflow_secrets:
  fernet_key: "v9IF2yPZtUH5aFnc6iGZsbBb3p5FpN6MemX4QWAHWPw="
  postgres_password: "random password"
  redis_password: "random password"
  airflow_ui_user_password: "random password"
  airflow_ui_user_email: "your.email@somehost.com"

!sensitive airflow_connections:
  mag_releases_table: "mysql://myname:mypassword@myhost.com"
  mag_snapshots_container: "mysql://myname:mypassword@myhost.com"
  crossref: "mysql://myname:mypassword@myhost.com"
  terraform: "mysql://:terraform-token"

airflow_variables:
  example_name: "example_value"
```

jdddog (commented):
Hey Aniek

Your changes are looking really good, that's a really good idea to use the yaml tags in the config files, that really enables them to be much more succinct.

Here are some small suggestions:
* Config file
  * It would be good if the connections (terraform, mag_releases_table, mag_snapshots_container and crossref) were optional. 
  * Rather than making the connections required, it would be good to validate the URI string, I'm not sure how easy that is, so don't worry if it looks too difficult. This is the code in airflow that makes the connections from the connection URI string: https://github.com/apache/airflow/blob/fdd9b6f65b608c516b8a062b058972d9a45ec9e3/airflow/models/connection.py#L202
  * In the config file generated by observatory generate terraform_config.yaml:
     * The !secrets tag should be added before google_application_credentials, airflow_secrets and airflow_connections in the autom.
     * workspace_prefix should be workspaces_prefix
     * airflow_worker_create: False should be airflow_worker_create: false
     * Some of the default values don't need to be commented out, e.g. database, airflow_main, airflow_main, airflow_worker_create, fernet_key.
     * Optional values, e.g. connections, could be commented out at the start of the line e.g. `# mag_releases_table: mysql://myname:mypassword@myhost.com`
     * It would be cleaner without the quoted strings where they are not necessary.
* The airflow local worker docker container doesn't start because the postgres password used to be [urlencoded in main.tf](https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/terraform/main.tf#L325). Perhaps this value should now be urlencoded in bash? As it also needed in the un-urlencoded form by resource "google_sql_user" "users".

Thanks

Jamie

aroelo:
> * Config file
>   
>   * It would be good if the connections (terraform, mag_releases_table, mag_snapshots_container and crossref) were optional.

Sure, I think I can hardcode this in somewhere in `create_terraform_schema`. 

>   * Rather than making the connections required, it would be good to validate the URI string, I'm not sure how easy that is, so don't worry if it looks too difficult. This is the code in airflow that makes the connections from the connection URI string: https://github.com/apache/airflow/blob/fdd9b6f65b608c516b8a062b058972d9a45ec9e3/airflow/models/connection.py#L202

Yes, that's a good idea. I actually already implemented something like that for the 'local' config file.
It's part of the ObservatoryConfigValidator class:
```
    def _validate_isuri(self, isuri, field, value):
        """ Validate that the given uri does not give an error when creating a Connection object. Will give a warning
        when uri doesn't contain a 'password' or 'extra' field.

        The rule's arguments are validated against this schema:
        {'type': 'boolean'}
        """

        if isuri:
            try:
                connection = Connection(uri=value)
                password = connection.password
                extra = connection.extra
                if not (password or extra):
                    print(f"WARNING: the uri in '{field}:{value}' does not contain a 'password' or 'extra' "
                          f"info.")
            except:
                self._error(field, f"the given uri '{value}' is not valid.")
```

I wasn't sure whether it should also check for as password, as maybe not all connections will always contain a password?
I could add the 'isuri' also somewhere in the `create_terraform_schema` function.

>   * In the config file generated by observatory generate terraform_config.yaml:
>     
>     * The !secrets tag should be added before google_application_credentials, airflow_secrets and airflow_connections in the autom.
>     * workspace_prefix should be workspaces_prefix
>     * airflow_worker_create: False should be airflow_worker_create: false
>     * Some of the default values don't need to be commented out, e.g. database properties, airflow_main properties, airflow_worker properties, airflow_worker_create, fernet_key.
>     * Optional values, e.g. connections, could be commented out at the start of the line e.g. `# mag_releases_table: mysql://myname:mypassword@myhost.com`
>     * It would be cleaner without the quoted strings where they are not necessary.

Thanks, these are all good points. I will try to implement them!

> * The airflow local worker docker container doesn't start because the postgres password used to be [urlencoded in main.tf](https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/terraform/main.tf#L325). Perhaps this value should now be urlencoded in bash? As it also needed in the un-urlencoded form by resource "google_sql_user" "users".

Ah, I overlooked this. I didn't run into any problems with this, probably because so far I've used 'test123' haha. Is it expected to have any special characters by convention? I saw that e.g the redis_password isn't urlencoded.

jdddog:
> Yes, that's a good idea. I actually already implemented something like that for the 'local' config file.
> It's part of the ObservatoryConfigValidator class:
> 
> ```
>     def _validate_isuri(self, isuri, field, value):
>         """ Validate that the given uri does not give an error when creating a Connection object. Will give a warning
>         when uri doesn't contain a 'password' or 'extra' field.
> 
>         The rule's arguments are validated against this schema:
>         {'type': 'boolean'}
>         """
> 
>         if isuri:
>             try:
>                 connection = Connection(uri=value)
>                 password = connection.password
>                 extra = connection.extra
>                 if not (password or extra):
>                     print(f"WARNING: the uri in '{field}:{value}' does not contain a 'password' or 'extra' "
>                           f"info.")
>             except:
>                 self._error(field, f"the given uri '{value}' is not valid.")
> ```
> 
> I wasn't sure whether it should also check for as password, as maybe not all connections will always contain a password?
> I could add the 'isuri' also somewhere in the `create_terraform_schema` function.

Yeah I think all of them will? They might not have a login/username though? E.g. an API token.

> > * The airflow local worker docker container doesn't start because the postgres password used to be [urlencoded in main.tf](https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/terraform/main.tf#L325). Perhaps this value should now be urlencoded in bash? As it also needed in the un-urlencoded form by resource "google_sql_user" "users".
> 
> Ah, I overlooked this. I didn't run into any problems with this, probably because so far I've used 'test123' haha. Is it expected to have any special characters by convention? I saw that e.g the redis_password isn't urlencoded.

Ah right, yeah they could have characters that are not valid for a url unless they are encoded.

jdddog:
I've tested the DAGs for turning the VM on and off and have added added a check to make sure that start_time_vm is not None, for when the VM hasn't been created at all: https://github.com/The-Academic-Observatory/observatory-platform/pull/231/commits/c86669634eca28d386e64c3b87ee615bbfab71c8#diff-b9cc1ba92da69da062889fc8c2473af2R177-R179. The DAGs are looking good, there are just a few things to look at.

* Add unit tests, especially for the TerraformApi class in [terraform_utils.py](https://github.com/The-Academic-Observatory/observatory-platform/blob/terraform_cloud_vm/observatory_platform/utils/terraform_utils.py).
* Fix up the failing unit tests.
* Something might not be quite right with how the vm_destroy DAG works (see below).

Something might not be quite right with how the vm_destroy DAG works, because after destroying the VM in the previous DAG run, in subsequent DAG runs, it runs all of the tasks without skipping, including applying a Terraform plan. Perhaps it is because the VM run date is set to the start_date of the DAG (e.g. dummy_telescope)?
```
[2020-09-07 23:50:15,844] {taskinstance.py:901} INFO - Executing <Task(ShortCircuitOperator): check_all_dags_success> on 2020-09-07T23:40:00+00:00
[2020-09-07 23:50:15,848] {standard_task_runner.py:54} INFO - Started process 275 to run task
[2020-09-07 23:50:15,892] {standard_task_runner.py:77} INFO - Running: ['airflow', 'run', 'vm_destroy', 'check_all_dags_success', '2020-09-07T23:40:00+00:00', '--job_id', '20', '--pool', 'default_pool', '--raw', '-sd', 'DAGS_FOLDER/vm_destroy.py', '--cfg_path', '/tmp/tmpdvppeyya']
[2020-09-07 23:50:15,893] {standard_task_runner.py:78} INFO - Job 20: Subtask check_all_dags_success
[2020-09-07 23:50:15,960] {logging_mixin.py:112} INFO - Running %s on host %s <TaskInstance: vm_destroy.check_all_dags_success 2020-09-07T23:40:00+00:00 [running]> airflow-main-vm.us-west1-c.c.academic-observatory-staging.internal
[2020-09-07 23:50:16,009] {terraform.py:174} INFO - Datetime when vm was turned on: 2020-09-07 23:22:14.122764+00:00
[2020-09-07 23:50:16,010] {terraform.py:181} INFO - Dag id: dummy_telescope
[2020-09-07 23:50:16,022] {terraform.py:190} INFO - First time running DAG. Setting last execution date to start_date of DAG.
[2020-09-07 23:50:16,022] {dagbag.py:396} INFO - Filling up the DagBag from /opt/airflow/dags
[2020-09-07 23:50:16,176] {terraform.py:197} INFO - Execution date of last run from previous VM cycle: 2020-08-10T00:00:00+00:00
[2020-09-07 23:50:16,218] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:34:00.177997+00:00, execution date: 2020-08-10 00:00:00+00:00, state: success
[2020-09-07 23:50:16,230] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:34:08.275377+00:00, execution date: 2020-08-11 00:00:00+00:00, state: success
[2020-09-07 23:50:16,241] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:34:16.372241+00:00, execution date: 2020-08-12 00:00:00+00:00, state: success
[2020-09-07 23:50:16,253] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:34:24.478213+00:00, execution date: 2020-08-13 00:00:00+00:00, state: success
[2020-09-07 23:50:16,264] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:34:32.584677+00:00, execution date: 2020-08-14 00:00:00+00:00, state: success
[2020-09-07 23:50:16,275] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:34:40.695174+00:00, execution date: 2020-08-15 00:00:00+00:00, state: success
[2020-09-07 23:50:16,285] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:34:48.791616+00:00, execution date: 2020-08-16 00:00:00+00:00, state: success
[2020-09-07 23:50:16,297] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:34:56.914318+00:00, execution date: 2020-08-17 00:00:00+00:00, state: success
[2020-09-07 23:50:16,308] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:35:05.007399+00:00, execution date: 2020-08-18 00:00:00+00:00, state: success
[2020-09-07 23:50:16,321] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:35:13.124815+00:00, execution date: 2020-08-19 00:00:00+00:00, state: success
[2020-09-07 23:50:16,332] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:35:21.297809+00:00, execution date: 2020-08-20 00:00:00+00:00, state: success
[2020-09-07 23:50:16,344] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:35:29.339393+00:00, execution date: 2020-08-21 00:00:00+00:00, state: success
[2020-09-07 23:50:16,356] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:35:37.435425+00:00, execution date: 2020-08-22 00:00:00+00:00, state: success
[2020-09-07 23:50:16,368] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:35:45.548983+00:00, execution date: 2020-08-23 00:00:00+00:00, state: success
[2020-09-07 23:50:16,379] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:35:53.652022+00:00, execution date: 2020-08-24 00:00:00+00:00, state: success
[2020-09-07 23:50:16,391] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:36:01.755086+00:00, execution date: 2020-08-25 00:00:00+00:00, state: success
[2020-09-07 23:50:16,403] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:36:09.862917+00:00, execution date: 2020-08-26 00:00:00+00:00, state: success
[2020-09-07 23:50:16,414] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:36:17.967652+00:00, execution date: 2020-08-27 00:00:00+00:00, state: success
[2020-09-07 23:50:16,424] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:36:26.071829+00:00, execution date: 2020-08-28 00:00:00+00:00, state: success
[2020-09-07 23:50:16,435] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:36:34.188449+00:00, execution date: 2020-08-29 00:00:00+00:00, state: success
[2020-09-07 23:50:16,446] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:36:42.283983+00:00, execution date: 2020-08-30 00:00:00+00:00, state: success
[2020-09-07 23:50:16,456] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:36:50.388002+00:00, execution date: 2020-08-31 00:00:00+00:00, state: success
[2020-09-07 23:50:16,467] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:36:58.507309+00:00, execution date: 2020-09-01 00:00:00+00:00, state: success
[2020-09-07 23:50:16,478] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:37:06.609782+00:00, execution date: 2020-09-02 00:00:00+00:00, state: success
[2020-09-07 23:50:16,489] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:37:14.727285+00:00, execution date: 2020-09-03 00:00:00+00:00, state: success
[2020-09-07 23:50:16,501] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:37:22.891402+00:00, execution date: 2020-09-04 00:00:00+00:00, state: success
[2020-09-07 23:50:16,512] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:37:30.925798+00:00, execution date: 2020-09-05 00:00:00+00:00, state: success
[2020-09-07 23:50:16,523] {terraform.py:221} INFO - id: dummy_telescope, start date: 2020-09-07 23:37:39.027585+00:00, execution date: 2020-09-06 00:00:00+00:00, state: success
[2020-09-07 23:50:16,523] {terraform.py:228} INFO - Destroying worker VM: True
[2020-09-07 23:50:16,524] {python_operator.py:114} INFO - Done. Returned value was: True
[2020-09-07 23:50:16,524] {python_operator.py:159} INFO - Condition result is True
[2020-09-07 23:50:16,524] {python_operator.py:162} INFO - Proceeding with downstream tasks...
[2020-09-07 23:50:16,538] {taskinstance.py:1070} INFO - Marking task as SUCCESS.dag_id=vm_destroy, task_id=check_all_dags_success, execution_date=20200907T234000, start_date=20200907T235015, end_date=20200907T235016
[2020-09-07 23:50:20,751] {local_task_job.py:102} INFO - Task exited with return code 0
```

observatory_platform/dags/dummy_telescope.py
```diff
@@ -0,0 +1,33 @@
...
+default_args = {
+    "owner": "airflow",
+    "start_date": datetime(2020, 8, 10)
+}
+
+
+with DAG(dag_id="dummy_telescope", schedule_interval="@daily", default_args=default_args, catchup=True) as dag:
+    telescope = DummyOperator(
+        task_id="telescope"
+    )
```
jdddog, line 31: Just add `queue='remote_queue'` to the DummyOperator so that when we use this DAG to test the VM create/destroy DAGs, the dummy_telescope DAG runs on the worker server.

jdddog: It seems that the queue parameter for `DummyOperator` gets ignored or it just runs on the default queue? So it  should be changed to the BashOperator, e.g. 

```
with DAG(dag_id="dummy_telescope", schedule_interval="@daily", default_args=default_args, catchup=True) as dag:
    task1 = BashOperator(
        task_id="task1",
        bash_command="echo 'hello'",
        queue='remote_queue'
    )

    task2 = BashOperator(
        task_id="task2",
        bash_command="echo 'world'",
        queue='remote_queue'
    )

    task1 >> task2
```

jdddog (approved):

PR #232, opened 2020-07-29
Restructured the way variables are loaded from config file.

aroelo:
Noticed when I wanted to add a 'terraform' connection inside Airflow I have to add this to 5 different methods of the ObservatoryConfig class. I wanted to make this process a little bit easier and this is just a proposal of one way to do it. You now only have to add it to one AirflowConn or AirflowVar class (besides changing code in other places which remains the same for both ways). 
What do you think @jdddog, do you have any suggestions?

Instead of hardcoding each individual airflow variable/connection as part of a config instance, each config instance now has two variables 'airflow_variables' and 'airflow_connections'. This makes it easier to add more variables/connections in the future.

To ensure that the right name for a variable/connection is used two classes were added: AirflowConn and AirflowVar. The variable/connection name from here is used when reading the config file, defining environment variables and inside telescopes.
This prevents errors regarding e.g. misspelling the variable/connection name.

Example config file:
```
environment: dev
data_location: us-east1
fernet_key: Jdl5louNP1
google_application_credentials: /Users/284060a/keys/gcp/workflows-dev-ead352480b42.json
airflow_connections:
  crossref: mysql://crossref-token@
  terraform: mysql://terraform-token@
  mag_releases_table: mysql://azure-storage-account-name:url-encoded-sas-token@
  mag_snapshots_container: mysql://azure-storage-account-name:url-encoded-sas-token@
airflow_variables:
  download_bucket_name: workflows-dev-910922156021-download
  transform_bucket_name: workflows-dev-910922156021-transform
  project_id: workflows-dev
```

### DIFF
--- a/observatory_platform/cli/observatory.py
+++ b/observatory_platform/cli/observatory.py
@@ -28,8 +28,8 @@ import docker
 import requests
 from cryptography.fernet import Fernet
 
-from observatory_platform.utils.config_utils import observatory_home, observatory_package_path, \
-    dags_path as default_dags_path, ObservatoryConfig
+from observatory_platform.utils.config_utils import AirflowVar, AirflowConn, observatory_home, \
+    observatory_package_path, dags_path as default_dags_path, ObservatoryConfig
 from observatory_platform.utils.proc_utils import wait_for_process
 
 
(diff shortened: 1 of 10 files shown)


observatory_platform/cli/observatory.py
```diff
@@ -158,15 +158,15 @@ def get_env(dags_path: str, data_path: str, logs_path: str, postgres_path: str,
...
 
     # Set variables
-    env['AIRFLOW_VAR_PROJECT_ID'] = config.project_id
+    env['AIRFLOW_VAR_PROJECT_ID'] = config.airflow_variables[AirflowVar.project_id.value]
     env['AIRFLOW_VAR_DATA_LOCATION'] = config.data_location
-    env['AIRFLOW_VAR_DOWNLOAD_BUCKET_NAME'] = config.download_bucket_name
-    env['AIRFLOW_VAR_TRANSFORM_BUCKET_NAME'] = config.transform_bucket_name
+    env['AIRFLOW_VAR_DOWNLOAD_BUCKET_NAME'] = config.airflow_variables[AirflowVar.download_bucket_name.value]
+    env['AIRFLOW_VAR_TRANSFORM_BUCKET_NAME'] = config.airflow_variables[AirflowVar.transform_bucket_name.value]
     env['AIRFLOW_VAR_ENVIRONMENT'] = config.environment.value
```
jdddog, line 170: Perhaps you could iterate over the airflow connections and variables here so that this section isn't hard coded too? Enums can be iterated: https://docs.python.org/3/library/enum.html#iteration and the name of the environment variable could be created with something like: `f'AIRFLOW_VAR_{enum_member.value.upper()}'`.

aroelo: That's a good idea, I've added this in the last commit (6e25630a5a0920b34081dcf433f85fde7e70881b) and made a couple of other small changes.

aroelo:
> Hey Aniek
> 
> These changes look good, just some of the tests that use the configuration files need to be updated:
> 
> * FAIL: test_platform_check_dependencies (tests.observatory_platform.cli.test_observatory.TestObservatory)
> * FAIL: test_platform_start_stop (tests.observatory_platform.cli.test_observatory.TestObservatory)
> * FAIL: test_from_dict (tests.observatory_platform.utils.test_config_utils.TestObservatoryConfig)
> * FAIL: test_is_valid (tests.observatory_platform.utils.test_config_utils.TestObservatoryConfig)
> * FAIL: test_make_default (tests.observatory_platform.utils.test_config_utils.TestObservatoryConfig)
> * FAIL: test_save (tests.observatory_platform.utils.test_config_utils.TestObservatoryConfig)
> * FAIL: test_to_dict (tests.observatory_platform.utils.test_config_utils.TestObservatoryConfig)
> 
> Thanks
> 
> Jamie

Updated the unit tests, they are now successful. Will work on some documentation tomorrow and rebase as well :-)

Issue #150, opened 2020-06-09
Update contributing instructions on CONTRIBUTING.md’


bechandcock:
Adding this as an issue to make sure it doesn’t get lost on Slack and other discussions.

The instructions on ‘CONTRIBUTING.md’ need to be updated. In their current form they are only helpful if one already knew which steps match with which phases, in which case one would not need to read instructions :-)

The section ‘**Developing a feature**’ is helpful, but the steps that follow in '**Detailed instructions on how to use Git ... '** do not match up in the numbering in ‘**Developing a feature’**. This makes it unclear which steps should be done prior to developing code, and which match up with updating the Academic Observatory with the code that you have developed.

Could the section starting '**Detailed instructions on how to use Git ... '** be re-numbered or divided into sections to match the three phases noted in ‘‘**Developing a feature**’? Or maybe some sub headings?

jdddog:
Thanks Rebecca, I'll have a look at this one this week

PR #241, opened 2020-08-27
Update CONTRIBUTING.md

jdddog:
This pull request refreshes the contributing guide:
* Added a table of contents and numbered headings.
* Split out documentation and unit tests into their own sections.
* Updated the following sections
  * Unit tests: how to test code that makes HTTP requests, updated information about where data should be stored.
  * License and copyright: described how to use the automatic license checker.
  * Development workflow: updated to reflect that we are no longer using forks and simplified the heading numbers so that the overview of the development workflow matches the details about how to use git https://github.com/The-Academic-Observatory/observatory-platform/issues/150


rhosking (approved):
Thanks Jamie, this is a really nice update. I suspect we will need to add more to this document as we go, but happy to merge as is

Issue #236, opened 2020-08-18
Crossref events missing required field 'source_token'


aroelo:
From their webpage (https://www.eventdata.crossref.org/guide/data/events/) the 'source_token' field is not optional. However, when I set the mode to "REQUIRED" accordingly in the schema, I get an error in BigQuery that this required field is missing.

Upon inspecting the data I can see there are a couple of entries without the 'source_token':
![image](https://user-images.githubusercontent.com/22207600/90461191-37466580-e138-11ea-8ed7-9cbeb7fc37ad.png)
<details> 

```
{"license": "https://creativecommons.org/publicdomain/zero/1.0/", "obj_id": "https://doi.org/10.15585/mmwr.mm6606a3", "source_token": "a6c9d511-9239-4de8-a266-b013f5bd8764", "occurred_at": "2017-02-17T07:14:20Z", "subj_id": "https://reddit.com/r/publichealth/comments/5ukuok/interim_estimates_of_201617_seasonal_influenza/", "id": "bb0af742-de56-4d7e-9d74-8d865e5a5e29", "evidence_record": "https://evidence.eventdata.crossref.org/evidence/201702238afe891c-3643-450b-9697-5a05b06044a3", "terms": "https://doi.org/10.13003/CED-terms-of-use", "action": "add", "subj": {"pid": "https://reddit.com/r/publichealth/comments/5ukuok/interim_estimates_of_201617_seasonal_influenza/", "type": "post", "title": "Interim estimates of 2016\u201317 seasonal influenza vaccine effectiveness \u2014 United States, February 2017 [CDC MMWR, 17 Feb 2017 \u2014 free full-text]", "issued": "2017-02-17T07:14:20.000Z"}, "source_id": "reddit", "obj": {"pid": "https://doi.org/10.15585/mmwr.mm6606a3", "url": "https://doi.org/10.15585/mmwr.mm6606a3"}, "timestamp": "2017-02-23T19:43:09Z", "relation_type_id": "discusses"}
[...]

Issue #78, opened 2020-05-22
Telescope workflow implementation: Crossref events


aroelo:
**Name**: Crossref events
**Subject Area**: Publication Events
**Harvest Type**: API 
**Query Type**: Pagination
**Snapshot Type**: N/A
**Snapshot Frequency**: N/A

See https://docs.google.com/spreadsheets/d/1ABHRxaGqjRq9j_oUftlSmhd2pBY09Pvc5n-JaleRbeI for more detailed info on data source.

aroelo:
Some small questions:
* It is not necessary, but preferred that we give an email address when using the crossref events API (https://www.eventdata.crossref.org/guide/service/query-api/#tell-us-who-you-are). So far I've used my curtin one; aniek.roelofs@curtin.edu.au. Which email address would be best to use in our production environment?

* When trying to get the data I often get a 503 error that the server is unavailable. What is a reasonable amount of retries before marking a task as failed in airflow? It's now set at 3 and tasks still fail quite regularly.

rhosking:
Your Curtin email address is fine for the moment, it is not super important. But I think once we move it into production, we should use a common email alias for things like this, that forwards emails to all of us (or the relevant sub group). I will take that on as a task to set something up.

Yeah, the 503 thing is interesting. This might be my lack of understanding of Airflow here, but is it possible to start exponentially backing off for each failure. Aka wait increasing longer amounts of time? The other thing, as we discussed, was there service was down for a few days. Is it possible for the task to just be retried at a later date? the backoff might achieve a similar result though

aroelo:
The 'until-updated' parameter is now working, see https://gitlab.com/crossref/issues/-/issues/877. The telescope should be updated accordingly.

PR #242, opened 2020-08-27
Telescope crossref events

aroelo:
Closing https://github.com/The-Academic-Observatory/observatory-platform/issues/236
Closing https://github.com/The-Academic-Observatory/observatory-platform/issues/78

The events are downloaded on a per day basis, either in parallel with the ThreadPoolExecutor or just sequentially.

The response is tried with 'retry' from tenacity so the download will be retried 3 times when the response status is not 200.
For some event days there are still issues with the API, and an internal server error might occur for all 3 retries. After the initial retry loop is exhausted it will try again, but with 100 rows per response (instead of the default max. of 1000 rows). This seems to fix the internal server error (it is suspected that the days that return an internal server error contain some large individual events which is the cause of the issue)

I've added some functionality to the bq_load methods of the StreamTelescope to load all the files that are in the download_folder into the table at once, using a wildcard (https://cloud.google.com/bigquery/docs/batch-loading-data#load-wildcards).
The files are not concatenated & compressed using gzip, because compressed files can't be loaded into BigQuery in parallel (so decompressed is much faster) and gzipped files can't be bigger than 4GB.

Also note that for March 2020 there is no event data available.


observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,701 @@
...
+        elif CrossrefEventsTelescope.DOWNLOAD_MODE == 'sequential':
+            batches = [(start_date, end_date)]
+        else:
+            raise AirflowException(f'Download mode has to be either "sequential" or "parallel", '
+                                   f'not "{CrossrefEventsTelescope.DOWNLOAD_MODE}"')
+
+        self.urls = []
+        edited_deleted = False
+        for batch in batches:
+            start_date = batch[0]
```
tuanchien, line 262: Not sure if you want to use this or not, but if you do, in the scopus branch there's a new class defined called SchedulePeriod that has .start and .end attributes of type pendulum.date for a little more human readability.

tuanchien (commented):
Good work Aniek!

I have left some minor comments.

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,701 @@
...
+        logging.info(f"{i + 1}.{j + 1} Downloading from url: {url}")
+        events_path = release.batch_path(url)
+        cursor_path = release.batch_path(url, cursor=True)
+        # check if cursor files exist from a previous failed request
+        if os.path.isfile(cursor_path):
+            # retrieve cursor
+            with open(cursor_path, 'r') as f:
+                next_cursor = json.load(f)
+                # delete file
+                pathlib.Path(cursor_path).unlink()
```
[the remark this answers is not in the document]
aroelo: Ah yes, I don't think this was on purpose. Thanks for picking this up, will move it outside the scope.

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,701 @@
...
+        """
+        start_date = self.start_date.date()
+        end_date = self.end_date.date()
+        max_processes = CrossrefEventsTelescope.MAX_PROCESSES
+
+        # number of days between start and end, add 1, because end date is included
+        total_no_days = (end_date - start_date).days + 1
+
+        # number of days in each batch, rounded down, minimal is 1
+        batch_no_days = max(1, round(total_no_days / max_processes))
```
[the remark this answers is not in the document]
aroelo: The comment is actually outdated from when I was previously rounding down, but now it is rounding to the closest integer. I will update the comment!

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,701 @@
...
+class CrossrefEventsTelescope:
+    """ A container for holding the constants and static functions for the crossref events telescope. """
+
+    DAG_ID = 'crossref_events'
+    DATASET_ID = 'crossref'
+    DESCRIPTION = 'The Crossref Metadata Plus dataset: ' \
+                  'https://www.crossref.org/services/metadata-retrieval/metadata-plus/'
+    RELEASES_TOPIC_NAME = "releases"
+    QUEUE = 'remote_queue'
+    MAX_PROCESSES = cpu_count()
```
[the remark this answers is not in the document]
aroelo: Thanks, I think I was confused with using processes instead of threads, similar to the discussion here https://stackoverflow.com/questions/40492894/python-multithreading-max-workers.

Do you have some tips on how to test whether it's more i/o bound than cpu bound? I'm actually not sure.

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,701 @@
...
+    MAX_CONNECTIONS = cpu_count()
+    MAX_RETRIES = 3
+    DOWNLOAD_MODE = 'parallel'  # Valid options: ['sequential', 'parallel']
+
+    MAILTO = 'aniek.roelofs@curtin.edu.au'
```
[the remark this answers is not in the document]
aroelo: It was intentional, since I was thinking we would replace it with a common email alias once this is created (see https://github.com/The-Academic-Observatory/observatory-platform/issues/78#issuecomment-691726153). 

I considered it similar to settings like download_mode, max_retries etc, however now that I think about it might be best to use an airflow connection/variable like you suggest, since we don't want others to use that e-mail when they clone the repository.

observatory_platform/telescopes/crossref_events.py
aroelo: Thanks, I'll have a look!

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,701 @@
...
+    :param url: The crossref events api url
+    :param events_path: Path to the file in which events are stored
+    :param next_cursor: The next cursor, this is in the response of the api
+    :param success: Whether all events were extracted successfully or an error occurred
+    :return: success, next_cursor and number of total events
+    """
+    headers = {
+        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1) AppleWebKit/537.36 (KHTML, like Gecko) '
+                      'Chrome/39.0.2171.95 Safari/537.36'
+    }
```
jdddog, line 67: For Crossref Events, it would be more appropriate to set a user agent that identifies our tool, so that we can be contacted and get better service. Having the mailto in the API query might be enough, but we might as well set an appropriate user agent as well. See the page here: https://github.com/CrossRef/rest-api-doc#etiquette

The user agent could be something like: `Observatory Platform (+https://github.com/The-Academic-Observatory/observatory-platform; mailto:agent@observatory.academy)`

observatory_platform/telescopes/crossref_events.py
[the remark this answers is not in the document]
jdddog: To quickly get an idea if the program is CPU bound you can use a program like `htop`.

To get an idea if it is i/o bound you could try `iostat` or `mpstat` and look at the field `%iowait`, if you have a very high percentage all the time then it might be i/o bound (it means that the CPU is waiting for disk requests).

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,701 @@
...
+        for x in range(no_batches - 1):
+            batch_end_date = batch_start_date + timedelta(days=batch_no_days - 1)
+            batches.append((batch_start_date.strftime("%Y-%m-%d"), batch_end_date.strftime("%Y-%m-%d")))
+            # end date is included, so start date should be one day later
+            batch_start_date = batch_end_date + timedelta(days=1)
+
+        # check if final batch has to be added
+        if batch_start_date <= end_date:
+            batches.append((batch_start_date.strftime("%Y-%m-%d"), end_date.strftime("%Y-%m-%d")))
+        return batches
```
jdddog, line 306: It might be simpler to use the Pendulum range function since we are already using Pendulum: https://pendulum.eustace.io/docs/#range

For example (from the Pendulum documentation): 
```python
import pendulum

start = pendulum.datetime(2000, 1, 1)
end = pendulum.datetime(2000, 1, 10)

period = pendulum.period(start, end)

for dt in period.range('days'):
    print(dt)
```

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,701 @@
...
+                        task_ids=CrossrefEventsTelescope.TASK_ID_DOWNLOAD, include_prior_dates=False)
+
+
+class CrossrefEventsTelescope:
+    """ A container for holding the constants and static functions for the crossref events telescope. """
+
+    DAG_ID = 'crossref_events'
+    DATASET_ID = 'crossref'
+    DESCRIPTION = 'The Crossref Metadata Plus dataset: ' \
+                  'https://www.crossref.org/services/metadata-retrieval/metadata-plus/'
```
jdddog, line 404: Should refer to Crossref Events rather than Crossref Metadata

jdddog (commented):
Hey Aniek

This is looking good, I've just left a few comments.

Thanks

Jamie

observatory_platform/telescopes/crossref_events.py
jdddog: Or you could set the mailto in the user agent as the same email you use in the mailto in the API calls

observatory_platform/telescopes/crossref_events.py
aroelo: I've used the pendulum range functionality and replaced it with this. I'm not sure if it is much simpler, might be a bit more readable.

```python
batches = []
period = pendulum.period(start_date, end_date)
for i, dt in enumerate(period.range('days', batch_no_days)):
    batch_start_date = dt.strftime("%Y-%m-%d")
    # if final batch or end date is reached before end of period, use end date of release instead
    if i == (max_processes - 1) or dt == end_date:
        batch_end_date = end_date.strftime("%Y-%m-%d")
        batches.append((batch_start_date, batch_end_date))
        break
    # use end date of period
    else:
        # end date is included, so subtract 1 day from batch_no_days
        batch_end_date = (dt + timedelta(days=batch_no_days - 1)).strftime("%Y-%m-%d")
        batches.append((batch_start_date, batch_end_date
```

observatory_platform/telescopes/crossref_events.py
aroelo: I've updated the user agent and included the mailto we use in the API calls :-)

observatory_platform/telescopes/crossref_events.py
```diff
@@ -477,10 +485,11 @@ def download(**kwargs):
         start_date = pendulum.instance(kwargs['dag_run'].start_date) - timedelta(days=1)
 
         logging.info(f'Start date: {prev_start_date}, end date:{start_date}, first release: {first_release}')
-        if prev_start_date < start_date:
+        if prev_start_date > start_date:
             raise AirflowException("Start date has to be before end date.")
 
-        release = CrossrefEventsRelease(prev_start_date, start_date, first_release)
+        mail_address = BaseHook.get_connection(AirflowConn.crossref_events.get()).password
```
tuanchien, line 491: If you want to use user/pass fields, may I suggest username instead for mail address. password might be harder to lookup once set due to airflow's security design

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,710 @@
...
+                batch_end_date = end_date.strftime("%Y-%m-%d")
+                batches.append((batch_start_date, batch_end_date))
+                break
+            # use end date of period
+            else:
+                # end date is included, so subtract 1 day from batch_no_days
+                batch_end_date = (dt+timedelta(days=batch_no_days-1)).strftime("%Y-%m-%d")
+                batches.append((batch_start_date, batch_end_date))
+
+        return batches
```
jdddog, line 314: It seems that the date range of each batch changes depending on what start_date and end_date is supplied. For instance, if a date range less than a month is supplied, it will create batches for each day, or if the date range is great enough it will split the batches into months. Would it make it simpler to just keep the batch period the same, except for the last period?

Tuan has created a function called build_schedule that is merged into develop, which you might be able to use here (or extend if you want slightly different functionality): https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory_platform/utils/telescope_utils.py#L36-L53

This type of functionality is good to put into its own function because it makes it easy to test on its own or reuse it.

observatory_platform/telescopes/crossref_events.py
```diff
@@ -0,0 +1,710 @@
...
+    :param events_path: Path to the file in which events are stored
+    :param next_cursor: The next cursor, this is in the response of the api
+    :param success: Whether all events were extracted successfully or an error occurred
+    :return: success, next_cursor and number of total events
+    """
+    headers = {
+        'User-Agent': 'Observatory Platform (+https://github.com/The-Academic-Observatory/observatory-platform; '
+                      f'mailto:{url.split("mailto=")[1].split("&")[0]})'
+    }
+    tmp_url = url + f'&cursor={next_cursor}' if next_cursor else url
```
jdddog, line 70: Tuan made a function `get_ao_user_agent` that has been merged with develop: https://github.com/The-Academic-Observatory/observatory-platform/blob/b62d87c07531f7ba35272393dca6da7734cf5f41/observatory_platform/utils/url_utils.py#L92

It might be a good idea to use and or extend this function to build the user agent string for standardisation purposes. You could add a keyword argument for the email address if you want to be able customise the email address (if the keyword argument email address is None then use the pkg_info email address).

jdddog (changes requested):
Hey Aniek

Nice work. There are just a couple of small things that I've made suggestions for in the comments.

Also, I the following error occurred when trying to install the dependencies (on Ubuntu):

```python
  File "/home/user/workspaces/aniek/observatory-platform/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 567, in _build_master
    ws.require(__requires__)
  File "/home/user/workspaces/aniek/observatory-platform/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 884, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/user/workspaces/aniek/observatory-platform/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 775, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (marshmallow 2.21.0 (/home/user/workspaces/aniek/observatory-platform/venv/lib/python3.7/site-packages), Requirement.parse('marshmallow>=3.0.0'), {'marshmallow-sqlalchemy'})
During handling of the above exception, another exception occurred:
```

@tuanchien gave me the following fix for this, which just involves adding these two packages to the requirements.txt:
* marshmallow==2.21.*
* marshmallow-sqlalchemy==0.23.*

Thanks

Jamie

observatory_platform/telescopes/crossref_events.py
aroelo: I preferred the use of flexible batch periods, so you optimise the use of number of processes available independent of the telescope's schedule. 

If we run the telescope every two weeks, we ensure that each process gets one batch of roughly the same size. And if we decide to change the schedule later on to e.g. weekly or monthly, each process still gets one batch of roughly the same size.

If the batch period is kept the same (for example one batch = one week), then this could be quite efficient when the schedule is set to bi-weekly, but if it is changed at one point to weekly or monthly, the batch size might be too small/big.

I think it comes down to how likely we'll think our schedules change, I thought a flexible batch size could account for this. If we're pretty set on keeping the schedule the same over time, it might be just overcomplicating things indeed!

I could add a variant of this (either with set batch period or flexible) to the telescope utils like Tuan did, I don't think the existing one is similar enough to what I'd need for crossref events (not using calender monthrange and I'd need a list of dates, instead of a list of periods)

observatory_platform/telescopes/crossref_events.py
aroelo: Thanks, I'll definitely add this.

aroelo:
> Hey Aniek
> 
> Nice work. There are just a couple of small things that I've made suggestions for in the comments.
> 
> Also, I the following error occurred when trying to install the dependencies (on Ubuntu):
> 
> ```python
> Traceback (most recent call last):
>   File "/home/user/workspaces/aniek/observatory-platform/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 567, in _build_master
>     ws.require(__requires__)
>   File "/home/user/workspaces/aniek/observatory-platform/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 884, in require
>     needed = self.resolve(parse_requirements(requirements))
>   File "/home/user/workspaces/aniek/observatory-platform/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 775, in resolve
>     raise VersionConflict(dist, req).with_context(dependent_req)
> pkg_resources.ContextualVersionConflict: (marshmallow 2.21.0 (/home/user/workspaces/aniek/observatory-platform/venv/lib/python3.7/site-packages), Requirement.parse('marshmallow>=3.0.0'), {'marshmallow-sqlalchemy'})
> During handling of the above exception, another exception occurred:
> ```
> 
> @tuanchien gave me the following fix for this, which just involves adding these two packages to the requirements.txt:
> 
> * marshmallow==2.21.*
> * marshmallow-sqlalchemy==0.23.*
> 
> Thanks
> 
> Jamie

Thanks for the review! I'll add these packages as well.

observatory-dags/observatory/dags/telescopes/crossref_events.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+class CrossrefEventsTelescope(StreamTelescope):
+    """ Crossref Events telescope """
+
+    DAG_ID = 'crossref_events'
+
+    MAILTO = 'aniek.roelofs@curtin.edu.au'
+    DOWNLOAD_MODE = 'parallel'  # Valid options: ['sequential', 'parallel']
+    MAX_PROCESSES = 21  # <ax processes based on 7 days x 3 url categories
```
jdddog, line 220: So that the email address used with Crossref Events can be customised, it would also be good if `mail_to` was a mandatory argument in the constructor. In the file `observatory-dags/observatory/dags/dags/crossref_events.py` you would then fetch the email address from an Airflow Connection.

It would also be good if `download_mode` and `max_processes` were keyword arguments in the constructor so that these can be customised when the telescope is created.

You might have to pass some of these values to the release object.

observatory-dags/observatory/dags/telescopes/crossref_events.py
```diff
@@ -0,0 +1,388 @@
...
+        if next_cursor:
+            success, next_cursor, total_events = extract_events(url, events_path, next_cursor, success)
+            return success, next_cursor, total_events
+        else:
+            return True, None, total_events
+    else:
+        raise ConnectionError(f"Error requesting url: {tmp_url}, response: {response.text}")
+
+
+def download_events_batch(release: CrossrefEventsRelease, i: int) -> list:
```
jdddog, line 333: This method might as well be part of the release object, since release is passed in as the first parameter.

observatory-dags/observatory/dags/telescopes/crossref_events.py
```diff
@@ -0,0 +1,388 @@
...
+        """ Task to transform the CrossrefEventsRelease release.
+
+        :param release: an CrossrefEventsRelease instance.
+        :return: None.
+        """
+        # Transform each release
+        release.transform()
+
+
+def extract_events(url: str, events_path: str, next_cursor: str = None, success: bool = True) -> \
```
jdddog, line 294: A name such as `download_events` might convey what this function does a little more.

observatory-dags/observatory/dags/telescopes/crossref_events.py
```diff
@@ -0,0 +1,388 @@
...
+            batch_start = url.split('from-collected-date=')[1].split('&')[0]
+            batch_end = url.split('until-collected-date=')[1].split('&')[0]
+        else:
+            batch_start = url.split('from-updated-date=')[1].split('&')[0]
+            batch_end = url.split('until-updated-date=')[1].split('&')[0]
+
+        if cursor:
+            return os.path.join(self.download_folder, f'{event_type}_{batch_start}_{batch_end}_cursor.txt')
+        else:
+            return os.path.join(self.download_folder, f'{event_type}_{batch_start}_{batch_end}')
```
jdddog, line 146: Perhaps this should have a file type of `.json`?

jdddog (changes requested):
Hey Aniek, it looks good, I've got a few comments for you. I'm running it now and might have a few more once it has finished.

observatory-dags/observatory/dags/telescopes/crossref_events.py
aroelo: Thanks, that makes sense. I've added them to the telescope and release objects.

observatory-dags/observatory/dags/telescopes/crossref_events.py
aroelo: Mmm, yes it did not mach up with the file type. I decided to keep it a .txt, but instead change the way the cursor was written to the file, so it is actually a text file. There was no need to use json.

jdddog (commented):
It looks good Aniek, it just needs to be rebased.

jdddog (approved):

Issue #21, opened 2020-04-29
Telescope workflow implementation: Web of Science


aroelo:
**Name**: Web of Science
**Subject Area**: Bibliometrics
**Harvest Type**: API
**Query Type**: DATE

See https://docs.google.com/spreadsheets/d/1ABHRxaGqjRq9j_oUftlSmhd2pBY09Pvc5n-JaleRbeI for more detailed info on data source.

PR #268, opened 2020-09-04
Telescope wos

tuanchien:
Design doc for the WOS source.

Particular points of interest:
 - naming conventions
 - the use of dynamic dags
 - whether to use dynamic tasks and partition the api retrieval task into separate tasks (either per call or per month or per year etc)
 - enforcement of similar bandwidth limits as scopus
 - schema design. this is a stripped down version of the old framework. the old schema seemed to merge msa,wos,scopus + some other meta information. i've only put in wos stuff as i didn't know what else you want in there.


tuanchien:
Linking issue #21

docs/telescopes/wos.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+## check_airflow [maybe this can be skipped]
+
+See if http://scientific.thomsonreuters.com is contactable.
+
+
+## download_data
+
+From the start date til the most recently completed month, pull data from the data source using an adapted version of Richard’s code, and append to a file on GC containing a list of json responses. Not sure if you want timing information like you did before.
```
jdddog, line 69: If we want to have a fresh copy of the most up to date data in Scopus for a particular institution, then we need to change the approach used to fetch data. The existing code only fetches data for a particular institution and month once, this means that fields such as the number of citations, will become out of date unless you fetch all of the records for each institution and date again, repeatedly.

So I think that we should change the architecture of how the telescope works to the following:
* The DAG runs every three months (could be a longer period, e.g. six months).
* Whenever the DAG runs, it fetches the records each institution, from a set date in the past (e.g. the year 2000) to the end date of the DAG run, even if it has fetched them before.
* All of the records fetched for each institution are given a release_date (the end date of the DAG run), which can be used to partition these records into a single release.

jdddog (commented):
Hey Tuan

This is looking good, I've put most of the comments in the review, here is a summary:
* The schema can be tweaked a little bit, by adding more specific field types.
* There are a lot more fields in WoS (and Scopus) that can be added to the schema, but it will need looking through the XML and Json example files to figure out how to represent them in BigQuery.
* I think we the DAGs to harvest the data for institutions, from a set date in the past to the end date of the DAG, repeatedly, otherwise the data will become out of date.

The same comments apply for Scopus.

Thanks

Jamie

docs/telescopes/wos.md
tuanchien: - I have concerns regarding the api quota if we do this backdating. Maybe @rhosking can share his experience on this.
- Regarding the set date for the institution, I was thinking of having that as a connection id attribute to minimise the amount of needed code changes.  Open to suggestions if you think that should be stored elsewhere.

tuanchien:
Changes:
 - uploaded a bq json schema with more fields than before.
 - removed field description in dataset_wos.md since Rebecca's project might generate this automatically in the future.
 - updated the doc description of wos to reflect suggestions on architecture and use of subdags from both wos and scopus threads.

tuanchien:
Ready for review again.  Todo: add descriptions to schema.

Updates:
 - Schema updated and a bunch of new fields added.
 - Dataset documentation left bare bones since it seems like autogeneration of documentation might be in the works.
 - Added utility functions to config_utils, gc_utils
 - Added functional implementation of wos telescope.
 - Separated out some functions into a telescope_utils file since some that will be reusable for scopus.
 - Default behaviour is to max out allowable sessions and parallelise fetching. A sequential mode is also available for conservative fetching.

Caveats:
 - Unclear how robust it is. Using 3rd party ratelimit and backoff packages.. It does currently have an effect, but unclear as to how robust without a lot of testing.  However, backoff/retry seems to work during throttle events.
 - Throttling limits are not optimised. Opted for lower limits than that specified by WoS.

jdddog (commented):
Hey Tuan

Your pull request is looking really good. I will need to run it to provide some more detailed feedback and to try out the subdags, I'll talk to you about that on Slack.

What do you think about making a class to represent a WoS Release? In some of the other telescopes we made a class that represents the particular release being processed and it wraps up the logic for making the file paths for the downloaded, extracted and transformed files. It also helps to simplify each task because you only need to send the list of Xcoms for a particular release once, from the download task and then that same topic is subscribed to from the rest of the tasks.

Thanks

Jamie

tuanchien:
I added a WosRelease class as per Jamie's comments above. It simplified some things, but download_paths didn't really simplify because the downloading is handled by WosUtility class, which has no access to pull messages from a taskinstance, which then would've required a taskinstance or release instance to be passed down the call chain anyway. Open to suggestions if you can think of a better way, or want to tempt the use of 'global' variables that are local to the .py file to get around that.

Other changes:
 - default mode changed to sequential so there's less hammering during testing.  can be changed to parallel in production maybe. if you want this as a configurable option in the connection extras field instead of as a class constant, let me know.
 - added WosUtilConst class to allow decorators to access them from within WosUtility method definitions
 - a few constants were cleaned up and made a class constant to cut down on a few more lines of code

jdddog (commented):
Hi Tuan

It would be good to add logging into each task (using the Python logging framework), especially for the download task, so that in the Airflow UI you can check the logs to see what queries have been run and what is going on in the other tasks.

Since the telescope is hitting an API endpoint, I think we should version the BigQuery schema files based on the WoS API version that is being used, rather than a date. A date doesn't make as much sense for an API as it does for a snapshot. Since if for an API, the DAG ran for a date in the past, it would use a schema version that might not be compatible with the latest API that was used.

What is the design decision behind using Pickle files? It would enable easier manual introspection if the raw XML was saved into the downloads folder rather than Pickle files.

It might be good to put the intermediate json files in the transformed directory and to not upload them, since they can be computed based on the input XML and they are really just there to make the final jsonl files that are uploaded in `upload_transformed`.

Thanks

Jamie

aroelo (commented):
Nice work overall! I didn't really look into the tests & code related to transforming/parsing the xml response, I might have time for this later after writing tests myself for my open PR :-).  Here's some of my questions/ideas that popped up while trying out the telescope:

* Is there supposed to be one table for each institution, or is this data merged? Initially I expected that there would be 1 subdag for each institution, and that each subdag would have tasks like download/upload_download/transform etc. This would give some more insight when a task fails at which institution the problem lies. It would also mean that if one institution fails, other institutions can independently run and their data can still be uploaded to their own table. 
* When I add an extra connection/institution, it seems the telescope processes only 1 of them. See for example the files in the 'download' directory (details) and in the buckets only files for 1 of the institutions are uploaded. (I had wos_curtin and wos_fake as connections, both using the curtin credentials you sent to me earlier)
<details>

```
-rw-r--r--  1 284060a  1930616786   7165078 18 Sep 12:49 curtin-2020-01-01_2020-09-18T04:49:02.203850+00:00.pkl
-rw-r--r--  1 284060a  1930616786   6460773 18 Sep 12:49 curtin-2020-02-01_2020-09-18T04:49:31.118853+00:00.pkl
-rw-r--r--  1 284060a  1930616786   5630517 18 Sep 12:50 curtin-2020-03-01_2020-09-18T04:49:56.146558+00:00.pkl
-rw-r--r--  1 284060a  1930616786   5492535 18 Sep 12:50 curtin-2020-04-01_2020-09-18T04:50:20.598662+00:00.pkl
-rw-r--r--  1 284060a  1930616786   5222941 18 Sep 12:51 curtin-2020-05-01_2020-09-18T04:50:40.361439+00:00.pkl
-rw-r--r--  1 284060a  1930616786   6558297 18 Sep 12:51 curtin-2020-06-01_2020-09-18T04:51:01.708056+00:00.pkl
-rw-r--r--  1 284060a  1930616786   4270309 18 Sep 12:51 curtin-2020-07-01_2020-09-18T04:51:23.961453+00:00.pkl
-rw-r--r--  1 284060a  1930616786   4363515 18 Sep 12:52 curtin-2020-08-01_2020-09-18T04:51:42.844528+00:00.pkl
-rw-r--r--  1 284060a  1930616786    872417 18 Sep 12:52 curtin-2020-09-01_2020-09-18T04:52:02.344165+00:00.pkl
-rw-r--r--  1 284060a  1930616786   6457209 18 Sep 12:58 fake-2020-01-01_2020-09-18T04:52:17.525038+00:00.json
-rw-r--r--  1 284060a  1930616786   1014994 18 Sep 12:58 fake-2020-01-01_2020-09-18T04:52:17.525038+00:00.json.gz
-rw-r--r--  1 284060a  1930616786   7165078 18 Sep 12:52 fake-2020-01-01_2020-09-18T04:52:17.525038+00:00.pkl
-rw-r--r--  1 284060a  1930616786   1044741 18 Sep 12:57 fake-2020-01-01_2020-09-18T04:52:17.525038+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786   5761512 18 Sep 12:58 fake-2020-02-01_2020-09-18T04:52:45.458141+00:00.json
-rw-r--r--  1 284060a  1930616786    913039 18 Sep 12:58 fake-2020-02-01_2020-09-18T04:52:45.458141+00:00.json.gz
-rw-r--r--  1 284060a  1930616786   6460773 18 Sep 12:53 fake-2020-02-01_2020-09-18T04:52:45.458141+00:00.pkl
-rw-r--r--  1 284060a  1930616786    937795 18 Sep 12:57 fake-2020-02-01_2020-09-18T04:52:45.458141+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786   5061398 18 Sep 12:58 fake-2020-03-01_2020-09-18T04:53:08.303943+00:00.json
-rw-r--r--  1 284060a  1930616786    797573 18 Sep 12:58 fake-2020-03-01_2020-09-18T04:53:08.303943+00:00.json.gz
-rw-r--r--  1 284060a  1930616786   5630517 18 Sep 12:53 fake-2020-03-01_2020-09-18T04:53:08.303943+00:00.pkl
-rw-r--r--  1 284060a  1930616786    817874 18 Sep 12:57 fake-2020-03-01_2020-09-18T04:53:08.303943+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786   4903356 18 Sep 12:58 fake-2020-04-01_2020-09-18T04:53:30.511439+00:00.json
-rw-r--r--  1 284060a  1930616786    765479 18 Sep 12:58 fake-2020-04-01_2020-09-18T04:53:30.511439+00:00.json.gz
-rw-r--r--  1 284060a  1930616786   5492535 18 Sep 12:53 fake-2020-04-01_2020-09-18T04:53:30.511439+00:00.pkl
-rw-r--r--  1 284060a  1930616786    785128 18 Sep 12:57 fake-2020-04-01_2020-09-18T04:53:30.511439+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786   4696942 18 Sep 12:58 fake-2020-05-01_2020-09-18T04:53:51.048906+00:00.json
-rw-r--r--  1 284060a  1930616786    739942 18 Sep 12:58 fake-2020-05-01_2020-09-18T04:53:51.048906+00:00.json.gz
-rw-r--r--  1 284060a  1930616786   5222941 18 Sep 12:54 fake-2020-05-01_2020-09-18T04:53:51.048906+00:00.pkl
-rw-r--r--  1 284060a  1930616786    756705 18 Sep 12:57 fake-2020-05-01_2020-09-18T04:53:51.048906+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786   5882224 18 Sep 12:58 fake-2020-06-01_2020-09-18T04:54:06.334873+00:00.json
-rw-r--r--  1 284060a  1930616786    911992 18 Sep 12:58 fake-2020-06-01_2020-09-18T04:54:06.334873+00:00.json.gz
-rw-r--r--  1 284060a  1930616786   6558297 18 Sep 12:54 fake-2020-06-01_2020-09-18T04:54:06.334873+00:00.pkl
-rw-r--r--  1 284060a  1930616786    934056 18 Sep 12:57 fake-2020-06-01_2020-09-18T04:54:06.334873+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786   3843785 18 Sep 12:58 fake-2020-07-01_2020-09-18T04:54:26.905794+00:00.json
-rw-r--r--  1 284060a  1930616786    598394 18 Sep 12:58 fake-2020-07-01_2020-09-18T04:54:26.905794+00:00.json.gz
-rw-r--r--  1 284060a  1930616786   4270309 18 Sep 12:54 fake-2020-07-01_2020-09-18T04:54:26.905794+00:00.pkl
-rw-r--r--  1 284060a  1930616786    613779 18 Sep 12:57 fake-2020-07-01_2020-09-18T04:54:26.905794+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786   3915047 18 Sep 12:58 fake-2020-08-01_2020-09-18T04:54:40.488243+00:00.json
-rw-r--r--  1 284060a  1930616786    620582 18 Sep 12:58 fake-2020-08-01_2020-09-18T04:54:40.488243+00:00.json.gz
-rw-r--r--  1 284060a  1930616786   4363515 18 Sep 12:54 fake-2020-08-01_2020-09-18T04:54:40.488243+00:00.pkl
-rw-r--r--  1 284060a  1930616786    635747 18 Sep 12:57 fake-2020-08-01_2020-09-18T04:54:40.488243+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786    786229 18 Sep 12:58 fake-2020-09-01_2020-09-18T04:54:54.019156+00:00.json
-rw-r--r--  1 284060a  1930616786    128967 18 Sep 12:58 fake-2020-09-01_2020-09-18T04:54:54.019156+00:00.json.gz
-rw-r--r--  1 284060a  1930616786    872417 18 Sep 12:54 fake-2020-09-01_2020-09-18T04:54:54.019156+00:00.pkl
-rw-r--r--  1 284060a  1930616786    131784 18 Sep 12:57 fake-2020-09-01_2020-09-18T04:54:54.019156+00:00.pkl.gz
-rw-r--r--  1 284060a  1930616786  22141595 18 Sep 12:56 fake_2020-08-01_2020-09-18T04:52:17.525038+00:00.jsonl
-rw-r--r--  1 284060a  1930616786   4752871 18 Sep 12:56 fake_2020-08-01_2020-09-18T04:52:17.525038+00:00.jsonl.gz

```
</details>

* I would add a subdirectory for each institution and DAG run date, or a combination of those, to store download/extract/transform files, so they are a bit more organised. 
* The transformed jsonl files should maybe be stored in the 'transform' subfolder, instead of 'download'
* You could possibly save some xcom usage by adding more info to the WosRelease. If the schedule becomes part of the WosRelease, I think you could derive the file names beforehand for zipped/json/blob files instead of getting the file names when executing these functions and then passing them as xcoms.
* If you do decide to use xcoms, maybe instead of pushing all xcoms to WosTelescope.RELEASES_TOPIC_NAME and having lists with dictionaries (e.g. `msgs_out = [{WosTelescope.XCOM_DOWNLOAD_PATH: file} for file in download_files]`) push the xcom directly as a list, using WosTelescope.XCOM_DOWNLOAD_PATH as the id. This can be done similarly for all of these:
```
    XCOM_DOWNLOAD_PATH = 'download_path'
    XCOM_UPLOAD_ZIP_PATH = 'download_zip_path'
    XCOM_JSON_PATH = 'json_path'
    XCOM_JSON_ZIP_PATH = 'json_zip_path'
    XCOM_HARVEST_DATETIME = 'harvest_datetime'
    XCOM_JSONL_PATH = 'jsonl_path'
    XCOM_JSONL_ZIP_PATH = 'jsonl_zip_path'
    XCOM_JSONL_BLOB_PATH = 'jsonl_blob_path'
```
I added an example in the review comments.
* The execution_date could be left out from the xcom_push usages, this is defaulted already to the execution_date.

tuanchien:
Updates as per review comments:
- More verbose logging added including session ids for downloaded files. A note on logging with subdags: you need to zoom in on the subdag and then click on individual tasks/logs to see the logs. The logs associated with the subdag itself will be scant.
- Schema file naming will follow the form: <prefix><tablename>_<version string>_<date schema added>.json
- Finding schema files: added "named schema" type which matches schema version string, and then uses the latest version.
- Pickle files removed in favour of xml text files. generates a lot more files, but might be easier to work with.
- json and jsonlines moved to transform directly
- json intermediate files not uploaded to cloud
- dag_start in download method changed to end_date to be more descriptive
- ETL pipeline moved into subdags. one per institution.
- Extra connection not working bug fixed by new subdag pipeline.
- New directory structure wos/<release date>/<institution>/timestamp.suffix
- provide_context set to False for check_api_server and kwargs removed in related method
- Added airflow variable overrides to help facilitate separation of tables / future acl implementation.
- Execution date removed from xcom push arguments
- XCOM messages push keys as topic names rather than nest under RELEASES_TOPIC_NAME 

New schema versioning
For WoS it seems to return a schema version in their response. For the time being I'm assuming that all query responses (irrespective of what records you fetch) will return their current version 'wok5.4', and I haven't seen any contrary responses so far (tested as far back as 1987). We have no guarantees though, and the documentation doesn't seem to be available so it's mostly what is observed from reverse engineering. The parsing code therefore assumes any wok5.4 response will be ok to process, and if this situation changes, the code may need changes.

However, we cannot solely rely on this version response from WoS because the field extraction is manual, and it doesn't seem to be possible to directly import the fields from raw json converted responses due to structural inconsistencies in responses. Since a subset of properties are extracted, there is scope for future extension so parsing output maybe different due to parsing changes.

The <version string>_<date schema added> setup tries to be capture both aspects. 

Release date
The release date is used to indicate when a snapshot was taken independent of the schema versioning.

Download schedules
The current design principle is to allow for possibility of dynamically optimising the number of api calls in the future. WoS returns a max of 100 results per call. Currently it's just pulling per month like the old behaviour but one could in the future analyse the distribution of results and try to minimise the number of api calls and see how much WoS can tolerate before it times out the session.

jdddog (changes requested):
Hey Tuan

The changes look good.

One thing to change, is that currently if you have keys for two institutions, the DAG that runs last, overwrites the table of the institution that was loaded into BigQuery previously. Perhaps to get around this, the load BigQuery task could run after all of the subdags have finished and load all of the jsonl files at once into the releases table, e.g. using the wildcard path syntax. The tables would also need to have a grid id column as well.

Thanks

Jamie

jdddog (dismissed):

jdddog (approved):

aroelo (approved):
Looks good :-) Nice usage of the 'backoff' and 'ratelimit' packages, I didn't know these exist and they might come in handy with some other telescopes as well.

PR #271, opened 2020-09-07
Telescope orcid

aroelo:


rhosking:
@aroelo has this branch/pull request superseded by #409?

aroelo:
> @aroelo has this branch/pull request superseded by #409?

Ah yep, I meant to push it to the same branch but accidentally did ORCID in lowercase instead of upper case. Could we keep it until the other one is merged? I might want to look back at the old code while updating the telescope with the template.

PR #274, opened 2020-09-24
Hotfix/installation

tuanchien:
- On some systems (mine, ubuntu 20.04), the installed version of importlib-metadata will be 2.0.0, whereas one of the package dependencies requires <2.  Since 1.7.4 works on Jamie's machine, targetting 1.4.* in requirements.txt
- Had weird artifact problem with airflow db where despite clearing out as many docker-compose images/orphans, and all docker images/containers and the config directory, the webserver container kept bootlooping due to missing table fields. Adding 'airflow initdb' to the entrypoint-airflow.sh script fixes this.  It's meant to be a non destructive command so it should be no harm no foul in adding it to handle that situation.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,3 +35,5 @@ ratelimit==2.2.*
 wos==0.2.*
 backoff==1.10.*
 oauth2client==1.5.2
+importlib-metadata==1.7.*
+


aroelo:
> * Had weird artifact problem with airflow db where despite clearing out as many docker-compose images/orphans, and all docker images/containers and the config directory, the webserver container kept bootlooping due to missing table fields. Adding 'airflow initdb' to the entrypoint-airflow.sh script fixes this.  It's meant to be a non destructive command so it should be no harm no foul in adding it to handle that situation.

I had to look into what the difference is between initdb/upgradedb, and am wondering how initdb fixes the missing table fields error. 
Here's a link to the source code: https://github.com/apache/airflow/blob/bcdd3bb7bb0e73ec957fa4077b025eb5c1fef90d/airflow/utils/db.py#L550. Could it be these two lines?

```python
from flask_appbuilder.models.sqla import Base
Base.metadata.create_all(settings.engine)  # pylint: disable=no-member
```

I don't understand what they do, but the other lines seem unrelated to missing table fields.

The entrypoint file is also used in our production environment I believe and it's not recommended to use initdb in production, because this creates all the default connections, see https://airflow.apache.org/docs/stable/best-practices.html?highlight=initdb#database-backend.

tuanchien:
The missing table stuff showed up in the logs as:

/home/airflow/.local/lib/python3.7/site-packages/airflow/configuration.py:664: DeprecationWarning: Specifying both AIRFLOW_HOME environment variable and airflow_home in the config file is deprecated. Please use only the AIRFLOW_HOME environment variable and remove the config file entry.
  warnings.warn(msg, category=DeprecationWarning)
[2020-09-23 23:31:09,774] {font_manager.py:1349} INFO - generated new fontManager
DB: postgres://airflow:***@postgres:5432/airflow
[2020-09-23 23:31:11,191] {db.py:378} INFO - Creating tables
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
/home/airflow/.local/lib/python3.7/site-packages/airflow/configuration.py:664: DeprecationWarning: Specifying both AIRFLOW_HOME environment variable and airflow_home in the config file is deprecated. Please use only the AIRFLOW_HOME environment variable and remove the config file entry.
  warnings.warn(msg, category=DeprecationWarning)
[2020-09-23 23:31:14,697] {manager.py:710} WARNING - No user yet created, use flask fab command to do it.
[2020-09-23 23:31:14,719] {base.py:592} ERROR - (psycopg2.errors.UndefinedTable) relation "ab_permission_view_role" does not exist
LINE 2: FROM ab_permission_view, ab_permission_view_role

This wasn't an issue from a clean vm, but on development environment it was erroring out even after clearing out the .observatory dir, and deleting all docker containers and images.

tuanchien:
See: https://stackoverflow.com/questions/57068702/sqlite3-raised-an-error-after-running-airflow-command-line for some discussion

aroelo:
Mmm alright. Don't know what's going on then. The only thing I can think of is that the airflow user is not created properly in the entrypoint file, I'm referring to this line:

```
# Create the Admin user. This command will just print "airflow already exist in the db" if the user already exists
  airflow create_user -r Admin -u airflow -e ${AIRFLOW_UI_USER_EMAIL} -f Observatory -l Admin -p ${AIRFLOW_UI_USER_PASSWORD}
```

As mentioned in the stackoverflow post the ab_* tables are for Role-based-access-control – RBAC, so they must be related to the created user somehow. Perhaps with initdb() the `Base.metadata.create_all(settings.engine)` creates this table regardless of airflow users? I'm just guessing here.

Do you have output in the webserver logs that the airflow users is created or already exists? And in the Airflow UI does it show user info?
image

Perhaps it is not that important to have default connections in the production environment and we should just include initdb if it solves the issue.

tuanchien:
If production is the main concern then I think we can probably omit initdb.  Am just suggesting we add it in if a similar situation arises with some funky artifacting with dev environments.  Since the lack of initdb doesn't error out a clean vm, it's unlikely to cause a problem in deployment.

importlib-metadata==1.7.* I think should be added though. That one does show up as an issue in a clean ubuntu 20.04 vm

tuanchien:
Regarding airflow ui/users, no webserver is launched because of that error, and docker forces it into a boot loop.
I didn't try adding a user before because I couldn't even get the airflow web ui.  But if it happens again I can try adding a new user first before initdb.

jdddog (approved):

PR #283, opened 2020-10-06
Update workspace-prefix name in create workspace test

aroelo:
Update workspace-prefix name to include unique datetime string when testing 'terraform create-workspace'.

If the workspace-prefix is not unique, it will give an error when running this workflow multiple times at the same time (e.g. pull request & push both trigger github workflow)

### DIFF
--- a/tests/observatory_platform/cli/test_observatory.py
+++ b/tests/observatory_platform/cli/test_observatory.py
@@ -20,12 +20,12 @@ import os
 import pathlib
 import sys
 import unittest
+from datetime import datetime
 from unittest.mock import patch
 
 from click.testing import CliRunner
 
-from observatory_platform.cli.observatory import cli, create_terraform_variables, gen_config_interface, \
-    gen_fernet_key, ObservatoryConfig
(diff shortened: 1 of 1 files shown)


tests/observatory_platform/cli/test_observatory.py
```diff
@@ -145,6 +144,7 @@ def test_platform_check_dependencies(self, mock_is_docker_running, mock_get_dock
 class TestObservatoryTerraform(unittest.TestCase):
     # TODO change to 'observatory-unit-testing' once latest version is available
     organisation = 'COKI-project'
+    workspaces_prefix = datetime.now().strftime("%Y-%M-%d_%H-%M-%S-%f")
```
jdddog, line 147: It might be a little safer to use a random uuid, it probably won't matter, but there is already a function `random_id` that will make one. You can import it with `from tests.observatory_platform.config import random_id`. It would at least make it consistent with the other tests that do similar things.

jdddog (commented):
This is great Aniek, I've just left one comment.

PR #284, opened 2020-10-06
Reports: Fixes to graphs and tests with related updates

cameronneylon:
A range of issues arose in the default naming of columns with
shifts to the new schema. These changes should not generally
be breaking and should actually fix problems that might not
have been previously obvious with the Output Types graph and
the funder graph.

I've also added a new test utility that will pull fresh test data so that this can be easily updated for
future schema changes. It also hard codes some of the publicness aspects that we adopted for the
eLife paper, i.e. only publicly naming universities that appear in the top 100.

This in turn led to a series of small changes to the tests. The tests remain less fully fleshed out than
they should be but they do all pass now.

Finally, I had to add precipy to requirements.txt to get tests to pass and took the opportunity to change
the seaborn version requirement to >=0.10.* New code in the groups report (not connected to this pull
request) will require >=0.11.0 but those are breaking changes which don't immediate effect anything here.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,9 +19,8 @@ tldextract==2.2.*
 urllib3==1.25.*
 matplotlib==3.1.*
 pandas==0.25.*
-seaborn>=0.10.*
+seaborn==0.11.*
 geopandas==0.8.*
-precipy>=0.2.3
 numpy==1.19.*
 ipython==7.16.*
 pydata-google-auth==1.1.*
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,8 +19,9 @@ tldextract==2.2.*
 urllib3==1.25.*
 matplotlib==3.1.*
 pandas==0.25.*
-seaborn==0.10.*
+seaborn>=0.10.*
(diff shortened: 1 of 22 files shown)


requirements.txt
```diff
@@ -19,8 +19,9 @@ tldextract==2.2.*
 urllib3==1.25.*
 matplotlib==3.1.*
 pandas==0.25.*
-seaborn==0.10.*
+seaborn>=0.10.*
```
jdddog, line 22: Perhaps set the requirement to `seaborn==0.11.*` so that we can get bug fixes but not minor / major version upgrades as Seaborn seems to have breaking changes even between minor version upgrades.

jdddog (commented):
That's great Cameron, there is just one minor comment about the version of Seaborn.

cameronneylon:
Hopefully final fixes now. Addressed the requirements.txt and bumped Seaborn to 0.11.* and removed precipy. Otherwise some minor adjustments and fixes that were necessary for the new group reports.

PR #286, opened 2020-10-12
Pbr setup

tuanchien:
Use Python pbr package for release management.
- generates version off git tags (or assigns one if missing)
- generates changelogs from commit log
- generates AUTHORS file from commit log
- sphinx integration for document generation

Added get_ao_user_agent() method in url_utils for a standardised user agent we can use in custom agents where we don't need to be sneaky.

The documentation generation stuff needs a closer look as I didn't do any testing on that.

As an aside, I noticed the following git commit duplications (pulled out by pbr):

Bec Handcock <43615088+bechandcock@users.noreply.github.com>
Bec Handcock <43615088+rnhandcock@users.noreply.github.com>
BecHandcock

Richard Hosking <richard.hosking@curtin.edu.au>
Richard Hosking <richard@Richards-MBP.hub>

aroelo <aniek.roelofs@curtin.edu.au>
aroelo <aniek93@live.nl>

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,3 +53,6 @@ tests =
     liccheck==0.4.*
     flake8==3.8.*
     coverage==5.2
+
+[pbr]
+skip_authors = true
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,55 @@
 [metadata]
+name = observatory-platform
+author = Curtin University
+author-email = agent@observatory.academy
+summary = The Observatory Platform is an environment for fetching, processing and analysing data to understand how well universities operate as Open Knowledge Institutions.
+    The Observatory Platform is built with Apache Airflow and includes DAGs (workflows) for processing: Crossref Metadata, Fundref, GRID, Microsoft Academic Graph (MAG) and Unpaywall.
+    See the Github Project here: https://github.com/The-Academic-Observatory/observatory-platform
+    See the documentation here: https://coki-academic-observatory.readthedocs-hosted.com/en/latest.
+    The Observatory Platform is compatible with Python 3.7.
+description = Telescopes, Workflows and Data Services for the Observatory Platform
 description-file = README.md
+description-content-type = text/markdown; charset=UTF-8
+home-page = https://github.com/The-Academic-Observatory/observatory-platform
+project_urls =
+    Bug Tracker = https://github.com/The-Academic-Observatory/observatory-platform/issues
+    Documentation = https://coki-academic-observatory.readthedocs-hosted.com/en/latest
+    Source Code = https://github.com/The-Academic-Observatory/observatory-platform
+    Download = https://github.com/The-Academic-Observatory/observatory-platform/v20.7.0.tar.gz
+license = Apache-2
+classifier =
+    Development Status :: 2 - Pre-Alpha
+    Environment :: Console
+    Environment :: Web Browser
+    Intended Audience :: Developers
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: Apache Software License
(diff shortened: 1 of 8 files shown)


tuanchien:
Ready for review.

If agreeable, we should use the standardised user agent string where it's safe for AO to present custom user agent strings.  This will impact some current pull requests. #269 and #242

setup.cfg
```diff
@@ -1,2 +1,58 @@
 [metadata]
-description-file = README.md
\ No newline at end of file
+name = observatory-platform
+author = Curtin University
+author-email = agent@observatory.academy
+summary = The Observatory Platform is an environment for fetching, processing and analysing data to understand how well universities operate as Open Knowledge Institutions.
+    The Observatory Platform is built with Apache Airflow and includes DAGs (workflows) for processing: Crossref Metadata, Fundref, GRID, Microsoft Academic Graph (MAG) and Unpaywall.
```
aroelo, line 6: I guess it's just copied from setup.py, but maybe we should leave out exactly which DAGs it includes or mention some as examples, since this will change frequently.

aroelo (approved):
Thanks Tuan, nice work.

Do you know where I would be able to sort out git commit duplications? Are these perhaps commits from my fork vs commits directly on this repository?

One question about the user agent, as I'm not sure what is common/best practice. Should it use the email address of the observatory platform for other users as well (since it was developed by the observatory platform) or should it be the email address of the actual user (e.g. the person/organisation that cloned the repository and is using it for themselves)?

setup.cfg
tuanchien: I guess an application of one of these things (wouldn't recommend filter branch though unless extreme care is taken):
https://stackoverflow.com/questions/2919878/git-rewrite-previous-commit-usernames-and-emails
I'm not entirely sure why those artifacts have shown up.  My guess would be that people maybe changed their local git configuration at some point.  I've disabled generation of AUTHORS file in the latest commit, so I guess it doesn't really even matter.

The email in user agent string is actually not very common, but if it is supplied, usually it is some bug reporting email for the project in case our crawler code is breaking something and the websites want to be angry at someone/report a bug. From that perspective it makes sense to supply a project email.  There is the edge case when a user downloads our code, breaks it, resulting in people still sending emails to us, but I'm not sure if that would actually be a big problem.

FYI, some example real bot user agents:
Mozilla/5.0 (Linux; Android 7.0;) AppleWebKit/537.36 (KHTML, like Gecko) Mobile Safari/537.36 (compatible; PetalBot;+https://aspiegel.com/petalbot)

- Mozilla/5.0 (compatible; Linux x86_64; Mail.RU_Bot/2.0; +http://go.mail.ru/help/robots)
- Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_5) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1.1 Safari/605.1.15 (Applebot/0.1; +http://www.apple.com/go/applebot)
- Mozilla/5.0 (Linux; Android 6.0.1; Nexus 5X Build/MMB29P) AppleWebKit/537.36 (KHTML, like Gecko) 
- Chrome/85.0.4183.140 Mobile Safari/537.36 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)
- Mozilla/5.0 (compatible; DotBot/1.1; http://www.opensiteexplorer.org/dotbot, help@moz.com)

Documentation note: great idea. Will make the change.

aroelo: Thanks for the explanation :-)

tuanchien: Made the documentation change. Need another review blessing :)

aroelo (approved):
🙏

PR #319, opened 2020-10-28
fix for funder aggregration

rhosking:
Small change, that was previous trialed in a different release, but the code was not committed to develop (sorry!)


jdddog (approved):
Looks good

Issue #84, opened 2020-05-24
Create a 'How to Write A Telescope' Guide


rhosking:
An opinionated flavour of the standard tutorials and guides for creating Airflow workflows, but catering towards the needs and context of the observatory platform.

I anticipate it to be as much of a hub for linking together all the best examples of best practice for Airflow more generally as it is about observatory specific practice. We don't want to diverge from the bulk on the community. Where we might need extra content is around the shared libraries we are creating for the types of data we are working with, alongside the continuing development of helper services that simply individual workflows

rhosking:
@jdddog @aroelo @tuanchien it would be good if you can all get together to discuss what this might look like once the current round of development starts wrapping up. With @jdddog proposed changes to the repositories, it makes sense to write this up once the dust settles

jdddog:
We can add @aroelo's suggestion here to the list: https://github.com/The-Academic-Observatory/observatory-platform/issues/225

aroelo:
Perhaps it would be nice to add a description that's visible in the UI for the tasks/DAG, see: https://airflow.apache.org/docs/stable/concepts.html#documentation-notes

jdddog:
> Perhaps it would be nice to add a description that's visible in the UI for the tasks/DAG, see: https://airflow.apache.org/docs/stable/concepts.html#documentation-notes

That would be great!

aroelo:
To better monitor the status of the DAGs, we could add an 'on_failure_callback' function to each DAG that will send a slack notification to the airflow channel (only in prod/staging?).

tuanchien:
It seems like there is a lot of duplicated work between telescopes.  For example, the airflow messaging, bq load, and cleanup.  Maybe we should implement template pipelines for most common patterns, and when people decide to use that, only get them to implement certain specialised things.

For example, download and transform might be the most commonly changed components, so if a simple telescope template got them to implement a download function that fetches, saves, and returns a list of files saved,  and a transform function that converts that into a json ready format, then you can save them from doing all the messaging and json conversion stuff.

tuanchien:
The concept of a Telescope needs to be more explicit.  Is it just anything that's not in the dag definition, responsible for all airflow and non airflow interactions, or should it be airflow specific interactions, with non airflow functionality offloaded to other functions classes.  Right now there seems to be a mixture where in some telescopes there is a mostly clean separation, with heavy lifting done in non telescope class methods in some cases, and a mix in other cases.

[...]

PR #321, opened 2020-11-09
Telescope template

aroelo:
I have added 3 telescope examples, based on the initial template ideas that I've made.
For all of these the idea is that the file where the DAG is created (dag.py) is minimally customised.
The only thing you should have to change is the telescope class name / import statement.

This way the dependencies structure is always the same for a specific type of telescope.
There are now 2 main types of telescopes, 'snapshots' and 'streams'.

**snapshot**
For the snapshot telescope a table shard is created for each snapshot in BigQuery.
Multiple snapshots can be processed in a single run, depending on how many snapshots are available between the start and end date of the telescope.
The start and end date correspond to the execution date and next execution date, catchup is set to True.

**stream** 
For the stream telescope two tables are created in BigQuery, 1 main table and 1 table with partitions (by ingestion date).
Only one update will be processed in a single run.
The start and end date are not determined by the execution date, but rather by the previous run date and current date, catchup is set to False.

**Customisable functions**
The functions that are characteristic of these telescopes, such as loading data into BigQuery or uploading files to a bucket should in principle not be customised, you could however overwrite them if this is required.

Examples of these functions are:
- upload_downloaded
- upload_transformed
- bq_load_shard
- bq_load_partition
- etc..

Some functions are required to customise, such as:
- list_releases
- download
- transfer
- download_transferred
- extract
- transform

And there are some that work without customisation, but you can optionally customise, such as:
- create_release
- check_dependencies (by default it will check for airflow connections and variables, but you could for example add an extra check specific for a connection, whether it is in the correct format)

**Examples**
For each example, I've added the minimal code, but also code that is commented out to give a more advanced example. It looks a bit messy, I'm not sure what the best way is to display both what is minimally required and an example of what is possible at the same time.

**Other types of telescopes**
These templates will work well for e.g. unpaywall, crossref metadata, crossref events, orcid and doab, but might not be fit for mag. I consider mag as a bit of an exception, since it has multiple snapshot tables for a single telescope. 
Similarly, for scopus and wos, it's slightly different, because these work with subdags. 
Depending on whether this will happen more often for telescopes we could make a separate category/template.

All feedback is much appreciated :-)

Closes https://github.com/The-Academic-Observatory/observatory-platform/issues/84
Closes https://github.com/The-Academic-Observatory/observatory-platform/issues/326


tuanchien:
Good work so far Aniek, I think there are some good ideas in there.  Might I suggest some further improvements in the way the code is organised to reduce the boiler plate burden.

So instead of remapping functions to *_custom internal names, we do a little bit of interfacing instead. Something like:

```python
from abc import ABC, abstractmethod

class Telescope(ABC):
""" Defines an interface for telescopes. """
    @classmethod
    def __subclasshook__(cls, subclass):
        return (hasattr(subclass, 'download') and callable(subclass.download) and
                    hasattr(subclass, 'extract') and callable(subclass.extract)
                or NotImplemented)
    
    @abstractmethod
    def download(self, **kwargs):
        """
        Download code.
        @param kwargs: Key arguments to downloader.
        """
        raise NotImplementedError

    @abstractmethod
    def extract(self, **kwargs):
        """
        Extract code.
        @param kwargs: Key arguments to extracter.
        """
        raise NotImplementedError
```
You could make specific interfaces for the different types of telescopes, or a generic one for all of them if you think there's a clean unified interface that would work. This interface is like a public contract with someone using it where we force them to implement those functions, in exchange, for predictable specified behaviour.

Then we would adopt this interface in a specialised telescope:
```python
class SpecialisedTelescope(Telescope):
  def __init__(self):
    # Actual work
    self.actual_variables = None

  def some_public_methods(self):
    return None

  def _private_method(self):
    return None
```

Then the user creates their own class that inherits the ```SpecialisedTelescopes```, but also implements ```download``` and ```extract``` methods, or those methods are implemented in the ```SpecialisedTelescopes``` and the user overrides them if they have different requirements.

The code in the interface class example makes sure that you cannot instantiate an object of any of those specialised classes without implementing ```download``` and ```extract```. Python interpreter will throw an error if you try.

If the machinery we build is specified well, users with more specialised requirements can implement their own classes that implement the defined interfaces, and re-use the machinery that relies on interfaces only, instead of specific interface implementations.

To cut down on boiler plate, we can have either a builder function or a builder class to construct the dag for us.

```python
class TelescopeDagFactory:
   def get_dag(name, parameters...):
       pass
```

In a dag file:
```python
dag = TelescopeDagFactory.get_dag(...)
dag.run()
```

or

```python
def get_dag(name, some parameters):
   construct_dag = None
   return constructed_dag
```

In a dag file:
```python
dag = get_dag(...)
dag.run()
```
or even just a
```python
def run_dag(name, parameters...):
   # Configure and run dag
   pass
```

This way you can just call those factory functions to return in custom dag files instead of copying and pasting a template in full.  Similarly, this builder/factory pattern can be used to configure and instantiate any telescope objects that don't actually require custom code but can just be parameterised.

aroelo:
> Good work so far Aniek, I think there are some good ideas in there. Might I suggest some further improvements in the way the code is organised to reduce the boiler plate burden.
> 
> So instead of remapping functions to *_custom internal names, we do a little bit of interfacing instead. Something like:
> 
> ```python
> from abc import ABC, abstractmethod
> 
> class Telescope(ABC):
> """ Defines an interface for telescopes. """
>     @classmethod
>     def __subclasshook__(cls, subclass):
>         return (hasattr(subclass, 'download') and callable(subclass.download) and
>                     hasattr(subclass, 'extract') and callable(subclass.extract)
>                 or NotImplemented)
>     
>     @abstractmethod
>     def download(self, **kwargs):
>         """
>         Download code.
>         @param kwargs: Key arguments to downloader.
>         """
>         raise NotImplementedError
> 
>     @abstractmethod
>     def extract(self, **kwargs):
>         """
>         Extract code.
>         @param kwargs: Key arguments to extracter.
>         """
>         raise NotImplementedError
> ```
> 
> You could make specific interfaces for the different types of telescopes, or a generic one for all of them if you think there's a clean unified interface that would work. This interface is like a public contract with someone using it where we force them to implement those functions, in exchange, for predictable specified behaviour.
> 
> Then we would adopt this interface in a specialised telescope:
> 
> ```python
> class SpecialisedTelescope(Telescope):
>   def __init__(self):
>     # Actual work
>     self.actual_variables = None
> 
>   def some_public_methods(self):
>     return None
> 
>   def _private_method(self):
>     return None
> ```
> 
> Then the user creates their own class that inherits the `SpecialisedTelescopes`, but also implements `download` and `extract` methods, or those methods are implemented in the `SpecialisedTelescopes` and the user overrides them if they have different requirements.
> 
> The code in the interface class example makes sure that you cannot instantiate an object of any of those specialised classes without implementing `download` and `extract`. Python interpreter will throw an error if you try.
> 
> If the machinery we build is specified well, users with more specialised requirements can implement their own classes that implement the defined interfaces, and re-use the machinery that relies on interfaces only, instead of specific interface implementations.

Thanks for the idea! I looked a little bit into abstract classes & methods as well when I was developing this, but I couldn't get it to work, because of the restriction that the telescope class can not be instantiated (or initialised? not sure about the terminology).

The methods of the SpecialisedTelescope class are ultimately passed on as a python callable to the PythonOperator, and at no point the class is instantiated. It would be nice if there's a way to work around this, but I haven't found it. Maybe @jdddog can confirm or help out.

 
> To cut down on boiler plate, we can have either a builder function or a builder class to construct the dag for us.
> 
> ```python
> class TelescopeDagFactory:
>    def get_dag(name, parameters...):
>        pass
> ```
> 
> In a dag file:
> 
> ```python
> dag = TelescopeDagFactory.get_dag(...)
> dag.run()
> ```
> 
> or
> 
> ```python
> def get_dag(name, some parameters):
>    construct_dag = None
>    return constructed_dag
> ```
> 
> In a dag file:
> 
> ```python
> dag = get_dag(...)
> dag.run()
> ```
> 
> or even just a
> 
> ```python
> def run_dag(name, parameters...):
>    # Configure and run dag
>    pass
> ```
> 
> This way you can just call those factory functions to return in custom dag files instead of copying and pasting a template in full. Similarly, this builder/factory pattern can be used to configure and instantiate any telescope objects that don't actually require custom code but can just be parameterised.

This is a great idea, thanks. I will try some things out, but if I understand correctly this could really help in cleaning up the DAG file itself.

jdddog (commented):
Nice work Aniek. I agree with Tuan's comments. Here are a few suggestions (one of them answers your question).

The classes that are created to represent different telescopes, should have explicit parameters in their constructors to enable customisation (see the example in the code).

If you have a function that constructs the actual DAG, you can just have one file in the dags folder that constructs all of the telescopes and makes the DAGs. The function `make_dag` has to return an Airflow `DAG` instance and in order for the DAG to get picked up by Airflow, it must be added to the globals symbol table, the DAG ID should be the key. This is illustrated below.

grid_telescope.py:
```python
from airflow import DAG

class GridTelescope:
  
  def make_dag(self):
      dag = DAG()

      with dag:
            task = PythonOperator()

     return dag
```

In the dags folder, load_dags.py:
```python
from datetime import datetime

from observatory.dags.telescopes.grid import GridTelescope

grid_telescope = GridTelescope(
  start_date=datetime(2020, 11, 1), 
  schedule_interval='@weekly', 
  catchup=True)
globals()[grid_telescope.dag_id] = grid_telescope.make_dag()
```

> The methods of the SpecialisedTelescope class are ultimately passed on as a python callable to the PythonOperator, and at no point the class is instantiated. It would be nice if there's a way to work around this, but I haven't found it. Maybe @jdddog can confirm or help out.

This is an example of what you could do:

```python
class SpecialisedTelescope(Telescope):
    
    def make_dag(self) -> DAG:
        dag = DAG()

         with dag:
             task = PythonOperator(
                    task_id='my_task_id',
                    provide_context=True,
                    python_callable=partial(self.__class__.task_callable, 'download')
                )

        return dag

    def download(self):
        " Download data" 
        pass

   @classmethod
    def task_callable(cls, func_name: str, **kwargs) -> None:
        # Get Airflow variables
        project_id = Variable.get(AirflowVars.PROJECT_ID)
        
        # Get Airflow connections
        my_conn = BaseHook.get_connection(AirflowConns.MY_CONN)
     
        # Create telescope instance
        instance = SpecialisedTelescope(
                       project_id=project_id,
                       my_conn=my_conn)

        # Make task function
        func = getattr(instance, func_name)

        # Call task function
        func()
```

This might seem a little in conflict with the example of the GridTelescope presented first, in terms of what parameters are set and where (in the load_dags.py file or in `task_callable`). It also doesn't have the concept of a dataset release which is quite common in our DAGs.

Perhaps instead of creating the `SpecialisedTelescope` instance within `task_callable`, you could pass the telescope instance to `task_callable` (which `make_dags` has access to) and then create a `Release` instance within `task_callable`. You can set all of the DAG run specific variables that you can only get access to when the DAG runs and `task_callable` is called (e.g. Airflow variables, connections, xcoms, DAG run start and end date etc). Then you might not need to use `getattr(instance, func_name)` etc to call the specific function on the `SpecialisedTelescope` instance. If that is not possible then you might need to create the `SpecialisedTelescope` instance and the `Release` instance within `task_callable`. Let me know if that makes sense or not?

For some more general feedback, I think it would be useful to pick a couple of the more complicated DAGs and backport them as you do this work (It would take too much time to do all of them), for example, the MAG telescope. The MAG telescope would be a good example because it doesn't neatly fit the idea of "download, extract, transform" (it transfers and then downloads), so it would help you to generalise your solution.

aroelo:
Thanks @tuanchien and @jdddog, both of your feedback really helped. I just committed a small updated example in a separate directory, in this commit: https://github.com/The-Academic-Observatory/observatory-platform/pull/321/commits/1ebeb7efdc3ea3214ab93de93d8a96d45d36c9db

I think this includes most of your ideas/tips, could you maybe give another round of feedback on this little bit before I refactor it in the same manner for the other methods/functions/templates etc?

aroelo:
> For some more general feedback, I think it would be useful to pick a couple of the more complicated DAGs and backport them as you do this work (It would take too much time to do all of them), for example, the MAG telescope. The MAG telescope would be a good example because it doesn't neatly fit the idea of "download, extract, transform" (it transfers and then downloads), so it would help you to generalise your solution.

Thanks for the review overall @jdddog. I'll just reply to this little bit, since I think the other things you mentioned are included in my last commit, thanks for those suggestions.

I do already have some functionality for 'transfer >> download' kind of telescopes, that would be the example in [dag.py](https://github.com/The-Academic-Observatory/observatory-platform/blob/1ebeb7efdc3ea3214ab93de93d8a96d45d36c9db/observatory_platform/templates/stream_transfer/dag.py) and [telescope.py](https://github.com/The-Academic-Observatory/observatory-platform/blob/1ebeb7efdc3ea3214ab93de93d8a96d45d36c9db/observatory_platform/templates/stream_transfer/telescope.py).

I will have a look at the MAG telescope, it can be considered a 'snapshot' telescope, but I expect it won't completely follow the template, as in, I will likely overwrite the 'bq_load' method, because it has to upload multiple tables to BigQuery for each snapshot. Normally, I would assume that there is only 1 table for each snapshot and the 'bq_load' method should not be overwritten.

I think it won't happen very often that there are multiple tables for a snapshot, so I wasn't planning on incorporating this into a template. What do you think?

observatory_platform/templates/updated_example/specialised_telescope.py
```diff
@@ -0,0 +1,130 @@
...
+        ti.xcom_push(self.RELEASES_XCOM, release)
+        ti.xcom_push(self.LAST_START_XCOM, end_date)
+
+    @abstractmethod
+    def download(self, release):
+        """
+        Download code.
+        :param release:
+        """
+        raise NotImplementedError
```
jdddog, line 102: The abstract methods, such as `download` should be defined in the `Telescope` class.

observatory_platform/templates/updated_example/specialised_telescope.py
```diff
@@ -0,0 +1,130 @@
...
+        """ Download release data to disk.
+
+        :param kwargs: The context passed from the PythonOperator.
+        :return: None.
+        """
+        ti: TaskInstance = kwargs['ti']
+        release = pull_release(ti)
+
+        success = self.download(release)
+        return True if success else False
```
jdddog, line 115: This method should override an abstract method from the `Telescope` class rather than calling an abstract method from the `Telescope` class.

observatory_platform/templates/updated_example/grid.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # def create_release(self, start_date: pendulum.Pendulum, end_date: pendulum.Pendulum, first_release: bool):
+    #     extra_param = 'extra'
+    #     release = GridRelease(self.dag_id, start_date, end_date, self.extensions, extra_param, first_release)
+    #     return release
+
+    def download(self, release) -> bool:
+        print('test from grid', self.dag_id)
+        print('download path & directory is created', release.download_path)
+        return True
```
jdddog, line 27: Ideally the `GridTelescope` would be the `SpecialisedTelescope`. I think that most of the functionality in the `SpecialisedTelescope` can be moved to `Telescope`.

jdddog (commented):
Hey Aniek

I think that the design of the `Telescope` abstraction needs a little more refinement. I've provided some comments in the code.

The [`ExportImportWorkflow`](https://github.com/The-Academic-Observatory/academic-observatory/blob/feature-elastic-integration/academic-observatory-dags/academic_observatory/dags/workflows/export_import_workflow.py) and [`ExportBigQueryImportElasticWorkflow`](https://github.com/The-Academic-Observatory/academic-observatory/blob/feature-elastic-integration/academic-observatory-dags/academic_observatory/dags/workflows/export_bigquery_import_elastic.py#L115) classes are good examples to look at.

The most useful aspect of this example is that it shows how to implement a general solution for a set of abstract methods that take a `Release` object as a parameter, which are called from a `PythonOperator` task. Each `PythonOperator` calls the method `task_callable`; with the abstract function that is desired to be executed supplied as a paramter (e.g.  `download_data` or `import_data`).  The `Release` object is created by calling `Release.make`. The supplied function is then called with the release as a parameter. Note that the `Release` class needs to be specialised, each type of release implements its own `make` function.

With this pattern the need for passing `Release` objects around with Xcoms can be mostly eliminated, since the same method, `Release.make` will be called to create the `Release` object, before every task method is called by `task_callable`. At most the date of the release needs to be pushed by the first task and then pulled in the `Release.make` method.

Some telescopes will need to be able to process multiple releases wont they? So then the interface should look more like:

```python
@abstractmethod
def download(self, releases: List[Release]) -> None:
```

The method in `Telescope` that creates the DAG instance, could change what tasks are created based on the parameters that are supplied to `Telescope`. 

The files in `observatory_platform/templates` and some of the files in `observatory_platform/utils` (including `telescope_snapshot.py`, `telescope_stream.py` and `telescope_utils.py`) can be moved into `observatory_platform/telescopes`. It is probably not necessary to have so many submodules as well (e.g. snapshot, stream_downloaded, stream_transfer etc). With the changes that have been merged into develop they would all ideally go in `observatory-platform/observatory/platform/telescopes` as it would be good if third party projects can use them as well without installing `observatory-dags`.

aroelo:
> Hey Aniek
> 
> I think that the design of the `Telescope` abstraction needs a little more refinement. I've provided some comments in the code.
> 
> The [`ExportImportWorkflow`](https://github.com/The-Academic-Observatory/academic-observatory/blob/feature-elastic-integration/academic-observatory-dags/academic_observatory/dags/workflows/export_import_workflow.py) and [`ExportBigQueryImportElasticWorkflow`](https://github.com/The-Academic-Observatory/academic-observatory/blob/feature-elastic-integration/academic-observatory-dags/academic_observatory/dags/workflows/export_bigquery_import_elastic.py#L115) classes are good examples to look at.
> 
> The most useful aspect of this example is that it shows how to implement a general solution for a set of abstract methods that take a `Release` object as a parameter, which are called from a `PythonOperator` task. Each `PythonOperator` calls the method `task_callable`; with the abstract function that is desired to be executed supplied as a paramter (e.g. `download_data` or `import_data`). The `Release` object is created by calling `Release.make`. The supplied function is then called with the release as a parameter. Note that the `Release` class needs to be specialised, each type of release implements its own `make` function.
> 
> With this pattern the need for passing `Release` objects around with Xcoms can be mostly eliminated, since the same method, `Release.make` will be called to create the `Release` object, before every task method is called by `task_callable`. At most the date of the release needs to be pushed by the first task and then pulled in the `Release.make` method.
> 
> Some telescopes will need to be able to process multiple releases wont they? So then the interfaces should look more like:
> 
> ```python
> @abstractmethod
> def download(self, releases: List[Release]) -> None:
> ```
> 
> The method in `Telescope` that creates the DAG instance, could change what tasks are created based on the parameters that are supplied to `Telescope`.
> 
> The files in `observatory_platform/templates` and some of the files in `observatory_platform/utils` (including `telescope_snapshot.py`, `telescope_stream.py` and `telescope_utils.py`) can be moved into `observatory_platform/telescopes`. It is probably not necessary to have so many submodules as well (e.g. snapshot, stream_downloaded, stream_transfer etc). With the changes that have been merged into develop they would all ideally go in `observatory-platform/observatory/platform/telescopes` as it would be good if third party projects can use them as well without installing `observatory-dags`.

Hey Jamie, thanks for the feedback!
I just had a look at the workflow examples and I think it works really well there, but I think there are some differences with the telescopes.

My view on the templates or standardisation is that we're not only creating 'helper' methods, but also determining the structure of a telescope and the dependencies between tasks. Is this what you had in mind as well?
This is why I think we need those 3 levels of telescopes (`Telescope`, `SpecialisedTelescope` and e.g. `Gridtelescope`), in the updated example `SpecialisedTelescope` is a bit like the `StreamTelescope`.

The `Telescope` (and `TelescopeRelease`) level can then contain all attributes/methods that are always required. An example might be 'check_dependencies'.

The `SpecialisedTelescope` level then contains all methods and attributes that are specific to that type of telescope. For now there are 2 types of telescopes (Stream and Snapshot) and I think it has added value to distinguish between these, because they are quite different to each other and most of the telescopes can be categorised as either one of the two. 

This way we can define the task dependencies inside the `SpecialisedTelescope` and use for example a ShortcircuitOperator for the right task (for streams this is when downloading a release, for snapshots this is when creating a list of releases). I prefer to not directly define the task callables as abstract methods, but instead a method called inside the task callable, because you can write some required code before or after that is required for each telescope.

For example download, for the 'stream' type there is always only one release, but for the 'snapshot' there might be multiple like you mentioned. In the snapshot telescope, the `download_task` function would be:
```python
    def download_task(self, **kwargs):
        """ Download all release data to disk.

        :param kwargs: The context passed from the PythonOperator.
        :return: None.
        """
        ti: TaskInstance = kwargs['ti']
        releases = pull_releases(ti)

        for release in releases:
            self.download(release)
```

The developer only has to write the download bit that will download a single release (and similarly for transform etc). 

Similarly for `check_dependencies`, you can have a task that already checks for the given airflow variables and connections inside the `check_dependencies_task`, but an additional `check_depencies` method that can be used to check for any extra dependencies. E.g. slightly updated:
```python
class Telescope(ABC):
    """ Defines an interface for telescopes. """
    def __init__(self, dag_id, start_date: datetime, schedule_interval, extensions: dict, airflow_vars: list,
                 airflow_conns: list):
        self.dag_id = dag_id
        self.start_date = start_date
        self.schedule_interval = schedule_interval
        self.extensions = extensions
        self.airflow_vars = airflow_vars
        self.airflow_conns = airflow_conns

    @abstractmethod
    def check_dependencies(self, kwargs):
        raise NotImplemented

    # a standard method that is shared by all different types of telescopes
    def check_dependencies_task(self, **kwargs):
        # check that vars and connections are available
        vars_valid = check_variables(self.airflow_vars)
        conns_valid = check_connections(self.airflow_conns)
        if not vars_valid or not conns_valid:
            raise AirflowException('Required variables or connections are missing')

        self.check_dependencies(kwargs)
```

Sorry if it's a bit confusing, to clarify, the 'stream_download', 'stream_transfer' and 'snapshot' are only given as examples. There are no seperate classes for transfer/download, but I thought I'd show that you can use either transfer or download for the two types (stream/snapshot) of telescopes.

It's a bit of a long comment, maybe it would be good to discuss some of these things on zoom?

observatory_platform/templates/updated_example/specialised_telescope.py
```diff
@@ -0,0 +1,130 @@
...
+
+    TASK_ID_CREATE_RELEASE = "create_release"
+    TASK_ID_DOWNLOAD = "download"
+    TASK_ID_BQ_LOAD_PARTITION = "bq_load_partition"
+
+    def __init__(self, dag_id, start_date, schedule_interval, extensions, partition_table_id):
+        super().__init__(dag_id, start_date, schedule_interval, extensions)
+        self.partition_table_id = partition_table_id
+
+    def create_airflow_DAG(self) -> DAG:
```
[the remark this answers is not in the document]
aroelo: The naming was on purpose, so that 'airflow' and 'DAG' are both mentioned whenever you create the DAG, to avoid this issue with airflow not finding the dag. (https://airflow.apache.org/docs/stable/faq.html) Maybe there's a better workaround, especially in the 'new' set-up with split repositories.

aroelo:
Closing as outdated. See new PR here: https://github.com/The-Academic-Observatory/observatory-platform/pull/359

PR #324, opened 2020-11-19
Hotfix group double counting

rhosking:
Addresses the problem where when a grid_id is matched against multiple groups, there was the risk of aggregate counts of outputs to be multiplied incorrectly


jdddog (approved):
We have tested this and it works

PR #325, opened 2020-11-19
Bugfix/scopuswos

tuanchien:
Scopus:
 - handle edge case of missing next links

Wos/get_as_list: 
 - Revert behaviour of get_as_list which was causing a regression. Return empty list instead of None.
 - Handle a couple of other 'None' edge cases.

get_ao_user_agent:
 - Switch to using importlib_metadata for package info.  pbr a bit flakey for grabbing that stuff.

### DIFF
--- a/observatory-dags/observatory/dags/telescopes/scopus.py
+++ b/observatory-dags/observatory/dags/telescopes/scopus.py
@@ -524,7 +524,7 @@ class ScopusClient:
 
             url = ScopusClient.get_next_page_url(response_dict['search-results']['link'])
             if url is None:
-                raise AirflowException(f'A missing links list was encountered while fetching results for {query}')
+                return results, quota_remaining, quota_reset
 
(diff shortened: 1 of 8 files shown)


observatory-dags/observatory/dags/telescopes/scopus.py
```diff
@@ -524,7 +524,7 @@ def retrieve(self, query):
```
jdddog, line 524: It would be good for this function (and others, it looks like most have type hints though) to use type hints to make it clear what the input and return types are, e.g. (not sure if these are correct).
```python
def retrieve(self, query: str) -> Tuple[List, int, int]:
```

jdddog (commented):
Hey Tuan

It looks good, the only thing that I can see is that some of the parameters and return types for methods are missing type hints.

jdddog (approved):
Looks good to me

PR #328, opened 2020-12-01
Update permissions granted to Google Cloud storage buckets by Terraform

jdddog:
This pull request updates the permissions that the Transfer Service Account and the Observatory Service Account are granted when accessing the Google Cloud storage buckets:

Transfer Service Account:
* Download bucket: storage.legacyBucketReader, storage.objectAdmin

Observatory Service Account:
* Download bucket: storage.legacyBucketReader, storage.objectCreator, storage.objectViewer
* Transform bucket: storage.legacyBucketReader, storage.objectAdmin
* Airflow bucket: storage.legacyBucketReader, storage.objectCreator, storage.objectViewer

I'm trying to not give the storage.objectAdmin permission if possible, especially to prevent accidental manual deletion of files in the download bucket.

The rationale for the Transfer Service Account having the storage.objectAdmin permission (rather than viewer and create) to the download bucket, is so that it can overwrite files if it needs to.

The rationale for the Observatory Service Account having storage.objectAdmin permissions to the transform bucket, is so that it can overwrite files if it needs to. This can happen if a DAG has a bug and transforms data incorrectly (which is uploaded to the transform bucket). Without the admin permission, when the DAG is run again and it uploads the newly transformed data, it will fail with a 403 error, as uploading the new files to the bucket requires deletion of the existing files. The files in this bucket get deleted automatically after 62 days anyway, so giving this permission isn't so bad.

What do you think?

### DIFF
--- a/observatory-platform/observatory/platform/terraform/main.tf
+++ b/observatory-platform/observatory/platform/terraform/main.tf
@@ -104,17 +104,10 @@ resource "google_project_iam_member" "observatory_service_account_storage_iam" {
   member  = "serviceAccount:${google_service_account.observatory_service_account.email}"
 }
 
-//module "google_application_credentials_secret" {
-//  source = "./secret"
-//  secret_id = "google_application_credentials"
-//  secret_data = google_service_account_key.observatory_service_account_key.private_key
-//  service_account_email = data.google_compute_default_service_account.default.email
-//  depends_on = [google_project_service.services]
-//}
-
 ########################################################################################################################
 # Storage Buckets
 ########################################################################################################################
+
 # Random id to prevent destroy of resources in keepers
 resource "random_id" "buckets_protector" {
   count = var.environment == "production" ? 1 : 0
@@ -156,7 +149,7 @@ resource "google_storage_bucket" "observatory_download_bucket" {
   }
 }
 
-# Permissions so that transfer service account can read / write files to bucket
+# Permissions so that the Transfer Service Account can read / write files to bucket
 resource "google_storage_bucket_iam_member" "observatory_download_bucket_transfer_service_account_legacy_bucket_reader" {
   bucket = google_storage_bucket.observatory_download_bucket.name
   role = "roles/storage.legacyBucketReader"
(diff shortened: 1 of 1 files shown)


aroelo:
I think this makes sense.

The only note I might add is that if the situation occurs where the data is transformed incorrectly, by the time you notice this it is likely that the data is already wiped from the disk locally (maybe not when developing, but in production).
This means that you'll already have to do some manual intervention by downloading the data from the bucket. Considering you already have this manual step, it might not be much effort to also manually delete the data from the bucket, this should all be contained in 1 file or folder specific to that run.

But like you said it might not add that much risk adding the permission anyway, and it will definitely make things easier in the case when the data is still on disk and you can just clear those Airflow tasks without having to manipulate the buckets manually.

tuanchien (dismissed):

jdddog:
> I think this makes sense.

Thanks Aniek.

> 
> The only note I might add is that if the situation occurs where the data is transformed incorrectly, by the time you notice this it is likely that the data is already wiped from the disk locally (maybe not when developing, but in production).
> This means that you'll already have to do some manual intervention by downloading the data from the bucket. Considering you already have this manual step, it might not be much effort to also manually delete the data from the bucket, this should all be contained in 1 file or folder specific to that run.
> 
> But like you said it might not add that much risk adding the permission anyway, and it will definitely make things easier in the case when the data is still on disk and you can just clear those Airflow tasks without having to manipulate the buckets manually.
> 

The times that I have had problems so far have been caused by transient errors that can't be replicated easily, e.g. something goes wrong with a transform task, so the transform and upload transformed data tasks have to be run again and once those are done the data loads into BigQuery OK.

> Thinking about this, another idea would be to let the role be dependent on the 'environment' variable, with only admin permission for the transform bucket in develop (and staging) environment.
> But this would assume that each environment per definition has its own google cloud project. I think this is how it is intended to work, but we might not be at that stage yet?

Yeah that is how it works, this is meant to address problems that occur in production though, so ideally the production bucket would have this permission.

tuanchien (approved):

PR #329, opened 2020-12-02
Github Actions workflow that publishes releases to PyPI

jdddog:
This pull request adds a Github Actions workflow, `publish-pypi.yaml`, that builds a tagged release of the Observatory and publishes it to PyPI. The action only runs when a tag is created.

The packages built and published include:
* observatory-platform: https://pypi.org/project/observatory-platform/
* observatory-dags: https://pypi.org/project/observatory-dags/
* observatory-reports: https://pypi.org/project/observatory-reports/

I've done a little bit of refactoring of the Github Actions workflows, namely, renamed pythonpackage.yml to unit-tests.yml (and called the workflow Unit Tests) as this workflow runs the unit tests.

This pull request is focusing on getting the workflow setup, so if there are suggestions for updating the documentation on PyPI or how we standardise the version numbering (this is independent of the workflow - it is just based on the Github tag) then that can be done in another ticket.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -54,9 +54,9 @@ jobs:
           packages_dir: observatory-dags/dist/
 
       - name: Publish observatory-reports
-          if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
-          uses: pypa/gh-action-pypi-publish@master
-          with:
-            user: __token__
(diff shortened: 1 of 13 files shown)


rhosking (approved):
thanks - looks good to me

rhosking (approved):
approved - as discussed, makes sense to put additional work to get this working correctly when installed from PyPi in another request

PR #337, opened 2020-12-15
Migrated dataquality

tuanchien:


tuanchien:
In case anyone wants to do a general review and leave comments or criticisms.

The unit tests are not passing for some reason.  Looks like I might've broken something in the reporting module with a version change to a dependency. Will take a look after new year.

PR #339, opened 2020-12-15
Update unpaywall schemas descriptions

aroelo:
Copied descriptions from https://unpaywall.org/data-format

Also re-ordered the fields, so they are in the same order as on the unpaywall webpage. 

I made sure to keep the differences between the different unpaywall versions.
For each of the older versions, these fields are excluded:
2020-02-25 vs 2020-10-06
- oa_locations.oa_date
- best_oa_location.oa_date
- first_oa_location

2019-11-22 vs 2020-02-25
- is_paratext

2018-09-24 vs 2019-11-22
- oa_locations.id
- best_oa_location.id


rhosking (approved):
Looks good to me - I think after all of these description changes are merged, a full run through of the whole system might be in order to catch any potential typos

PR #350, opened 2021-01-28
Initial proposal for Unpaywall aggregation query from COKI Sprint #1

cameronneylon:
Following the recommendations of the Sprint #1 group the query has been
modified to use a CASE WHEN syntax for readability and has been
heavily commented.

Following the findings of the group on improved precision in the Unpaywall
data, slight modifications to the logic of the queries have been made
particularly with respect to Bronze determination.

Further additional categories will be the subject of future Sprints and
the documentation is only a first pass to investigate how best to manage
and record this.

All changes are to aggregate_unpaywall.sql.jinja2

### DIFF
--- a/observatory-dags/observatory/dags/database/sql/aggregate_unpaywall.sql.jinja2
+++ b/observatory-dags/observatory/dags/database/sql/aggregate_unpaywall.sql.jinja2
@@ -12,7 +12,41 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Author: Richard Hosking, James Diprose, Contributors #}
+# Author: Richard Hosking, James Diprose, Contributors
+
+## AGGREGATE UNPAYWALL QUERY TEMPLATE
+
+This template query contains the SQL that directly interprets Unpaywall
+data to determine OA categories at the output level. This is therefore
+the canonical location for the precise definitions used for OA categories.
+Ideally this file should contain both the queries themselves and
+a description clear enough for a non-expert in SQL to understand how each
+category is defined.
+
+The current categories of Open Access described in this file are:
+
+* is_oa: derived directly from Unpaywall
+* hybrid: accessible at the publisher with a recognised license
+* bronze: accessible at the publisher with no recognised license
+* gold_just_doaj: accessible at the publisher site from a journal in DOAJ
+* green: accessible at any site recognised as a repository (including preprints)
+* green_only: accessible at a repository and not (gold or hybrid or bronze)
+* green_only_ignoring_bronze: accessible at a repository and not (gold or hybrid)
+
+### DEVELOPER NOTES and DOCUMENTATION GUIDANCE
+
+All changes to OA categories should be made in this file and no changes should be
+made elsewhere. Each category defined in the SELECT statement should be
+in the form of a CASE-WHEN statement that returns TRUE or FALSE (this
+may be changed in the future to 1 or 0).
+
+Documentation should seek to provide a clear explanation of the category
+including any differences from common usage to a non-expert user. It is
+intended that this documentation will provide the public-facing
+version of the categories and should always be edited in close proximity
+to the queries themselves. Strive for clarity and precision over
+brevity where necessary
+#}
 
 SELECT
   UPPER(TRIM(doi)) as doi,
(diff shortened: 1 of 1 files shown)


rhosking:
Initial impressions are that I like the use of the CASE statement, it does help with readability. Documentation is also nice.

Once suggestion for the documentation, and I don't think this is clear cut at all, but it might be better to wrap those comments up in jinja2 template comments, so when they are executed, the query logs in the system won't be as long.

The other comment, as this applies equally to the current state of things, and this new set of changes, is how we might want to write tests, perhaps this a sample dataset to ensure this change, and future changes don't break anything going forwards?

cameronneylon:
### Documentation

Happy to be guided on the comments. I was torn whether to put them at the top where (I imagined) they would get pulled into the docstring or having them close to the actual code. It would be ideal if they could be pulled into readthedocs in the same way the docstrings for a function are on a per-SELECT statement basis but I realise that doesn't actually make any sense, even if readthedocs understood SQL. 

The Sprint Group didn't have a specific view on how the comments should work, except that they would ideally be in one canonical place, as close as practicable to the actual code and also available, ideally automatically, in more general-user friendly form. Overall this strikes me a general challenge for us in being transparent as we can through the whole stack so its a good place to think it through. 

### Tests

We did do some testing in the background to make sure things work but a more rigorous approach would make sense. My suggestion would be a test fixture that can be used as example data where we know exactly what the answers should be, derived from a recent Unpaywall snapshot so we can also do proper internal checks (nothing both hybrid and bronze for instance).

Is there a way of testing the "contract" with the downstream aggregation workflow? I can imagine we want to know that the generated table has all the necessary columns. This might be worth thinking about now as I expect there will be another round of proposed changes which may be breaking changes in terms of rationalising the naming as we add new categories.

rcoscom:
I also appreciate more the clarity of the CASE statements. I think the OA tagging reads much clearer now. Maybe, the only difference I would suggest is to work with 1 (for TRUE cases) and 0 (for FALSE cases) to allow for potential calculations based on the labelling (e.g. weighting OA by co-authorships, by fields, etc.).

cameronneylon:
Thanks for that Rodrigo - I might record that as an issue for the next round. I'm conscious in this round of not making any changes to the data that flows downstream. I *think* it would still work with the downstream code, but would prefer to test that separately :-)

rhosking:
The status of this pull request is still set to draft. Just checking as to whether it is ready for the final review before hopefully being merged?

Agree about thinking deeper about testing, but I'm not sure it's essentially to have that all solved before merging this set of changes. Measuring any effects of any data/sql feeding data into the workflows and how that may alter any downstream dataset is high on my desired set of automated tests

cameronneylon:
I’m waiting on a little more input at this point, hopefully in the next few days.

Cameron Neylon
Professor of Research Communications
Centre for Culture and Technology
School of Media, Creative Arts and Social Inquiry
Curtin University

bmkramer:
Thanks for consolidating this Cameron! 

A couple of points, partly also me trying to wrap my head around the details at micro-level ;-) 

line 73-77: 
`CASE
    WHEN journal_is_in_doaj OR (best_oa_location.host_type = "publisher" AND best_oa_location.license is not null AND not journal_is_in_doaj) THEN TRUE
    ELSE FALSE
  END
    as gold,`
    
-> is `AND not journal_is_in_doaj` superfluous here, since it's an OR relation so all `journal_is_in_doaj` is included? Or is this done to combine the definitions for gold_just_doaj and hybrid in the definition of gold?

line 141-146 / 155-160, as well as 31, 33, 34
` WHEN (SELECT COUNT(1) FROM UNNEST(oa_locations) AS location WHERE location.host_type IN ('repository')) > 0 AND
      NOT (journal_is_in_doaj OR best_oa_location.host_type = "publisher")`
  This excludes articles in DOAJ-journals that, despite that, are only available via a repository (n=150 in the entire database). That's indeed fine if for gold_just_doaj, journal title is taken as defining variable, not the actual availability via the publisher (as indeed is done in line 85-89. In that case though, shouldn't it also be reflected in the description of gold_just_doaj? (i.e. should the 'accessible at the publisher site' be removed in favor of 'articles from a journal in DOAJ'?)  
  
  line 28-34
  A description of the category 'gold' is missing
  
  line 96 (commenting)
  `the use of DOAJ as defining a "fully oa journal" is also narrow and future developments will
expand this.` - add: `considering, among other parameters, the Unpaywall tag 'journal-is-oa'. `?

line 138-139 / 152-153 (commenting)
`This category is mainly used for the generation of stacked bar charts that include gold_doaj,
green, hybrid and bronze.` - suggestion to broaden the potential application a bit: `This category enables analyses of gold(+bronze) and green as mutually exclusive categories, e.g. in the generation of stacked bar charts that include gold_doaj,
green, hybrid and bronze`

line 162-166 / 168-172
Add commenting, to prevent potential confusion (that these lines are part of the definition of green_only_excluding_bronze), and also, to make clear that this is a designation for the license for best_oa_location only?

cameronneylon:
> line 73-77:
CASE WHEN journal_is_in_doaj OR (best_oa_location.host_type = "publisher" AND best_oa_location.license is not null AND not journal_is_in_doaj) THEN TRUE ELSE FALSE END as gold,

> -> is AND not journal_is_in_doaj superfluous here, since it's an OR relation so all journal_is_in_doaj is included? Or is this done to combine the definitions for gold_just_doaj and hybrid in the definition of gold?

Yes, this was written so that the definition was effectively gold_just_doaj OR hybrid. I'm inclined to leave it that way as it makes it clearer what we're doing when we potentially expand "pure" gold definitions?

> line 141-146 / 155-160, as well as 31, 33, 34
WHEN (SELECT COUNT(1) FROM UNNEST(oa_locations) AS location WHERE location.host_type IN ('repository')) > 0 AND NOT (journal_is_in_doaj OR best_oa_location.host_type = "publisher")
This excludes articles in DOAJ-journals that, despite that, are only available via a repository (n=150 in the entire database). That's indeed fine if for gold_just_doaj, journal title is taken as defining variable, not the actual availability via the publisher (as indeed is done in line 85-89. In that case though, shouldn't it also be reflected in the description of gold_just_doaj? (i.e. should the 'accessible at the publisher site' be removed in favor of 'articles from a journal in DOAJ'?)

Agreed. I will keep the old logic and note this as an issue for the future. There are a non-trivial number of cases where Unpaywall has missed one or two articles in an apparently fully-OA journal so this will need to be a future policy decision on characterisation. This will apply to both cases where we are trying to detect OA journals (substantial number with just one non-OA article) as well as cases where a journal is in DOAJ but should no longer be (presumably not a massive problem if you only found 150 articles in that category?)

At line 31, 33, 34 have updated the documentation to make this clear.

For the rest of the comments I have taken them on board and added the relevant documentation in 2e0706b

cameronneylon:
This is good to go from my perspective at this point. It does need an overview check, to check numbers don't change too much (or really at all), and I think that would be better run from the file itself, rather than me cutting and pasting out the SQL from the jinja but otherwise good to go.

cameronneylon:
Old Query:

Row | total | oa | gold | hybrid | bronze | green | green_only | gold_doaj | green_only_ignoring_bronze | has_license | cc_licensed |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 118469729 | 34619355 | 10533095 | 4785178 | 16170250 | 16291635 | 7916115 | 5747917 | 14815925 | 11457713 | 10481454

New Query:

Row | total | oa | gold | hybrid | bronze | green | green_only | gold_doaj | green_only_ignoring_bronze | has_license | cc_licensed
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 118469729 | 34619355 | 10533095 | 4785178 | 16170250 | 16291635 | 7916010 | 5747917 | 11284265 | 11457713 | 10481454

The difference in "green only ignoring bronze" is expected and not very important

observatory-dags/observatory/dags/database/sql/aggregate_unpaywall.sql.jinja2
```diff
@@ -12,31 +12,182 @@
...
+    as is_cclicensed,
+
+{#
+### Provides a list of repository URLs for future analysis by repository type
+#}
+  ARRAY((SELECT url FROM UNNEST(oa_locations) WHERE host_type = "repository")) as repository_locations,
   best_oa_location.url_for_landing_page,
   best_oa_location.url_for_pdf
+
+FROM `academic-observatory.our_research.unpaywall20201009`
```
rhosking, line 190: this line needs to be removed

rhosking (changes requested):
Just doing a quick comparison between the old version of the query and the updated one here. The results are the following, each field being counting the number of times a value is different between the two versions:


Row | oa | is_in_doaj | has_license | is_cclicensed | gold_just_doaj | green | gold | hybrid | green_only | green_only_ignoring_bronze | bronze |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 105 | 3531660 | 0

They all match, aside from the green and green_only_ignoring_bronze. Was that expected?

the list of dois for each class that don't match:
- coki-scratch-space:richard.dois_of_green_not_matching
- coki-scratch-space:richard.dois_of_green_only_ignoring_bronze_not_matching

observatory-dags/observatory/dags/database/sql/aggregate_unpaywall.sql.jinja2
bmkramer: The difference in green_only is as expected 
(for my own reference as much as anyone else's:  the new query counts all articles in DOAJ journals as gold, also when Unpaywall labels them as only available from a repository)

The difference in green_only_ignoring_bronze is also indeed as expected 
(@cameronneylon the new query fixed an OR vs AND mistake, right? It now correctly excludes articles that are either gold_in_doaj *or* hybrid, which is correct))

cameronneylon:
Yes, that’s correct the previous “green_only_ignoring_bronze” was incorrect so the difference is expected. Should be minimal dependencies as I’m not aware of that value actually being used anywhere at the moment in any case.

cameronneylon:
Corrected the extraneous FROM line in the most recent push so the pull request should be good to go by Monday

rhosking (approved):
Looks good - happy to merge

PR #369, opened 2021-03-02
Geonames Telescope template port

jdddog:
This pull request ports the Geonames telescope to use the new telescope template and uses the Observatory test environment for unit testing.


observatory-platform/observatory/platform/telescopes/snapshot_telescope.py
```diff
@@ -54,19 +59,28 @@ def __init__(self, dag_id: str, start_date: datetime, schedule_interval: str, da
         :param queue: the Airflow queue name.
         :param max_retries: the number of times to retry each task.
         :param max_active_runs: the maximum number of DAG runs that can be run at once.
+        :param source_format: the prefix used to find the schema path
```
aroelo, line 62: Description of schema_prefix instead of source_format

tests/observatory/dags/telescopes/test_geonames.py
```diff
@@ -45,28 +40,46 @@ def __init__(self, *args, **kwargs):
...
+            'check_dependencies': ['list_releases'],
+            'list_releases': ['download'],
+            'download': ['upload_downloaded'],
+            'upload_downloaded': ['extract'],
+            'extract': ['transform'],
+            'transform': ['upload_transformed'],
+            'upload_transformed': ['bq_load'],
+            'bq_load': ['cleanup'],
+            'cleanup': []
+        }, dag)
```
aroelo, line 66: What do you think about using the actual references to these methods? It doesn't look as nice as the strings, but might prevent extra work later if we for example change the method name 'bq_load' in the telescope template.

```
        dag = GeonamesTelescope().make_dag()
        t = GeonamesTelescope
        self.assert_dag_structure({
            t.check_dependencies.__name__: [t.list_releases.__name__],
            t.list_releases.__name__: [t.download.__name__],
            t.download.__name__: [t.upload_downloaded.__name__],
            t.extract.__name__: [t.transform.__name__],
            t.transform.__name__: [t.upload_transformed.__name__],
            t.upload_transformed: [t.bq_load.__name__],
            t.bq_load.__name__: [t.cleanup.__name__],
            t.cleanup.__name__: []
        }, dag)
```

tests/observatory/dags/telescopes/test_open_citations.py
```diff
@@ -46,6 +47,7 @@ def test_open_citations_release(self, mock_variable_get):
         """
 
         # Mock data variable
+        reset_variables()
```
aroelo, line 50: When do you need to use the 'reset_variables()' function?

aroelo (commented):
Really nice, it will be great to be able to use the test environment for the telescopes :-)

I saw you added some functionality to the bigquery related methods of the template, thanks for adding those and fixing some other things up in the template as well.

I have a couple of questions/suggestions and added them as comments in the code.

observatory-dags/observatory/dags/telescopes/geonames.py
```diff
@@ -110,266 +59,196 @@ def first_sunday_of_month(datetime: Pendulum) -> Pendulum:
...
 
-        # Get variables
-        bucket_name = Variable.get(AirflowVars.DOWNLOAD_BUCKET)
+        assert len(
+            release_dates) == 1, f"There should only be one Geonames release: len(release_dates) is {len(release_dates)}"
 
-        # Upload file
-        upload_file_to_cloud_storage(bucket_name, release.get_blob_name(SubFolder.downloaded),
-                                     file_path=release.filepath_download)
+        return [GeonamesRelease(self.dag_id, release_dates[0])]
```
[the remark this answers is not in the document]
jdddog: That's true, I don't know what I was thinking haha

tests/observatory/dags/telescopes/test_open_citations.py
jdddog: We won't need to use it once all of the telescopes are using the telescope template and the observatory test environment.

It is being used because we use global variables in [template_utils.py](https://github.com/The-Academic-Observatory/observatory-platform/blob/feature-geonames-telescope-template/observatory-platform/observatory/platform/utils/template_utils.py#L49-L54) so that the Google Cloud Secrets backend isn't called too much when requesting certain Airflow Variables. For these global variables, once the Airflow Variable has been fetched once, then it won't be fetched again, see [here](https://github.com/The-Academic-Observatory/observatory-platform/blob/feature-geonames-telescope-template/observatory-platform/observatory/platform/utils/template_utils.py#L83-L86) for an example. The problem when running unit tests, is that the first test to run which sets one of these global variables, will end up setting that value for all of the unit tests (as they won't update after that). When the tests run, the functions in template_utils.py need to be able to get the latest values for the Airflow Variables, which is why reset_variables() needs to be called before these tests are run.

These global variables are automatically reset in the observatory test environment, so it won't be a problem once we use that.

observatory-dags/observatory/dags/telescopes/geonames.py
```diff
@@ -110,266 +59,196 @@ def first_sunday_of_month(datetime: Pendulum) -> Pendulum:
...
+        if continue_dag:
+            # Fetch release date
+            release_date = fetch_release_date()
 
-    @staticmethod
-    def transform(**kwargs):
-        """ Transform release into a jsonl file.
+            # Push messages
+            ti: TaskInstance = kwargs['ti']
+            ti.xcom_push(GeonamesTelescope.RELEASE_INFO, [release_date], execution_date)
```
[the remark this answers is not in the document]
jdddog: I've renamed this to `fetch_release_date`.

jdddog:
> Really nice, it will be great to be able to use the test environment for the telescopes :-)
> I saw you added some functionality to the bigquery related methods of the template, thanks for adding those and fixing some other things up in the template as well.

Thanks Aniek!

> One more question that came to mind, where do the unit tests get the credentials for the Google Cloud project from? I could not find it that easily

The Google credentials are set with the Github secret: TESTS_GOOGLE_APPLICATION_CREDENTIALS_BASE64, which is decoded and written as a file in the Github Actions environment, the path to this file is set with the environment variable GOOGLE_APPLICATION_CREDENTIALS: https://github.com/The-Academic-Observatory/observatory-platform/blob/feature-geonames-telescope-template/.github/workflows/unit-tests.yml#L55-L56

When you run it on your own machine it also uses the GOOGLE_APPLICATION_CREDENTIALS environment variable.

observatory-platform/observatory/platform/telescopes/snapshot_telescope.py
jdddog: Thanks, I've fixed that now.

observatory-dags/observatory/dags/telescopes/geonames.py
```diff
@@ -110,266 +59,196 @@ def first_sunday_of_month(datetime: Pendulum) -> Pendulum:
...
+                         catchup=catchup, airflow_vars=airflow_vars)
+        self.add_setup_task(self.check_dependencies)
+        self.add_setup_task(self.list_releases)
+        self.add_task(self.download)
+        self.add_task(self.upload_downloaded)
+        self.add_task(self.extract)
+        self.add_task(self.transform)
+        self.add_task(self.upload_transformed)
+        self.add_task(self.bq_load)
+        self.add_task(self.cleanup)
```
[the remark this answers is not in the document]
jdddog: > I noticed the construction of Telescopes is slightly different for the telescopes and even though it's a small detail I think it would be nice to be consistent.
> 
> * In this case there is a DAG_ID class variable which is passed on to the dag_id parameter.
> * The DATASET_DESCRIPTION is a class variable, but passed on to the dataset_description parameter in the super call .
> * airflow_vars is a variable in the init method (rather than being a parameter)
> * Some values are directly defined in the super() init call, like 'source_format' and 'load_bigquery_table_kwargs'. Rather than being parameters in the child init or (class) variables.
> 
> When constructing a telescope for the dag (see below), which properties of the telescope do we want to be able to configure during this process?
> 
> ```
> from observatory.dags.telescopes.grid import GridTelescope
> 
> telescope = GridTelescope()
> globals()[telescope.dag_id] = telescope.make_dag()
> ```
> 
> In this example you could create the telescope and fill in the start_date, schedule_interval, dataset_id and catchup, but not the dag_id, dataset_description, airflow_vars or bigquery_table_kwargs. It seems that in practice (so far) we don't configure any of these, but leave them to the default values.

Yeah it would be good to make it consistent.

We will want to be able to customise the dag_id parameter, because we will create multiple DAG instances (with different ids) with the new Mellon telescopes, based on what the REST API returns.

The class variable DAG_ID (as opposed to self.dag_id) is still required for all telescopes as they are used by the [doi.py DAG](https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory-dags/observatory/dags/dags/doi.py#L73-L108).

> 
> If we will always leave them to the default values I would prefer to list the parameters as variables in the init method (like airflow_vars), just because it looks a bit cleaner to me having them below each other compared to on 1 line as parameters.
> 
> I'm just wondering what the reasoning is behind which properties are parameters/which are not and if we should make this consistent across telescopes.

My thinking was that the users wouldn't need to change certain parameters in the final telescopes (e.g. GeonamesTelescope), e.g. source_format, load_bigquery_table_kwargs, airflow_vars etc. I'm happy to make them all keyword arguments in the constructor if that makes them all consistent though.

tests/observatory/dags/telescopes/test_geonames.py
jdddog: I personally like to use strings because it is independent of anything defined in the telescope class, so the test forces you to make sure that the DAG structure is correct if there are changes.

tuanchien (approved):

tests/observatory/dags/telescopes/test_open_citations.py
aroelo: Ah okay, thanks for the explanation!

observatory-dags/observatory/dags/telescopes/geonames.py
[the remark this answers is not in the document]
aroelo: > The class variable DAG_ID (as opposed to self.dag_id) is still required for all telescopes as they are used by the doi.py DAG.

Thanks, I did not realise this, but good to know.

> My thinking was that the users wouldn't need to change certain parameters in the final telescopes (e.g. GeonamesTelescope), e.g. source_format, load_bigquery_table_kwargs, airflow_vars etc. I'm happy to make them all keyword arguments in the constructor if that makes them all consistent though.

Sounds good :-)

PR #376, opened 2021-03-12
Feature Organisation & Telescope API

jdddog:
This pull request adds API endpoints to manage:
* Organisations, including their Google Cloud project information (project id, bucket names etc).
* Telescopes, including what telescopes each organisation has and the specific type of each telescope (so that they can be filtered).

I've split the API out into it's own Python package called `observatory-api`, so that it can easily be built independently of the rest of the Observatory into it's own Docker container and also tested in the same tests folder as the other packages.

The API can be previewed and experimented with here: https://endpointsportal.coki-jamie-dev.cloud.goog/docs/coki-jamie-dev.api.observatory.academy/1/introduction

I will generate a Python client to use this API and show an example of how to use the API to create Airflow DAGs in a future pull request (the ONIX telescope).


tuanchien (approved):
Looks good to me

observatory-platform/observatory/platform/terraform/main.tf
```diff
@@ -347,6 +353,13 @@ resource "google_compute_firewall" "allow_ssh" {
   priority = 65534
 }
 
+resource "google_vpc_access_connector" "observatory_vpc_connector" {
+  name = local.vpc_connector_name
+  ip_cidr_range = "10.8.0.0/28"
+  network = local.network_name
+  region = var.google_cloud.region
+}
+
```
aroelo, line 362: Could you maybe elaborate a bit on what the vpc access connector does? It seems like this connects the cloud run backend to the vpc network, is this required for the added features of the API?

rhosking (dismissed):
looks nice! looking forward to thinking about the next steps for these features

aroelo (commented):
Nice work Jamie!

I think the example would help to understand how this added functionality can be used for a telescope like ONIX. I have a general idea, but it is still a bit unclear to me, so looking forward to that. 
One of the questions I have that might become clear with the example is about the telescope type. What is the purpose of this? It seems like it has the same properties as a telescope (id, name, created, modified).
Would it be possible to add extra information to a telescope/telescope type? An example I'm thinking of is for the google analytics telescope, it requires a view ID that is different for each organisation. We could also get this with e.g. airflow variables, but just thinking whether storing it in this database could be a solution.

The only comment I have is that the readme and installation.md need a minor update with info to optionally install the observatory-api.

I like the restructuring that you did of the files and the render template fix, so we don't need to maintain 2 openapi configurations :-)

observatory-platform/observatory/platform/terraform/main.tf
jdddog: The VPC Access Connector is required to enable the Cloud Run backend to connect to the CloudSQL database. I can add it as a comment.

jdddog:
> Nice work Jamie!
> 

Thanks!

> I think the example would help to understand how this added functionality can be used for a telescope like ONIX. I have a general idea, but it is still a bit unclear to me, so looking forward to that.
> One of the questions I have that might become clear with the example is about the telescope type. What is the purpose of this? It seems like it has the same properties as a telescope (id, name, created, modified).

TelescopeType defines the types of telescopes that can be created, e.g. the ONIX Telescope, Google Books Telescope, Google Analytics Telescope etc.

Telescope defines the telescope instances that exist, e.g. the Curtin University Press ONIX Telescope, which should then be created by Apache Airflow.

The TelescopeType is used to return all of the instances of a specific type of telescope so that they can all be created at once.

> Would it be possible to add extra information to a telescope/telescope type? An example I'm thinking of is for the google analytics telescope, it requires a view ID that is different for each organisation. We could also get this with e.g. airflow variables, but just thinking whether storing it in this database could be a solution.

Possibly, ideally the properties would be general enough for all organisations or all telescopes, perhaps we can talk about this offline to figure out how to do it?

> The only comment I have is that the readme and installation.md need a minor update with info to optionally install the observatory-api.

Thanks, I'll update this.

> I like the restructuring that you did of the files and the render template fix, so we don't need to maintain 2 openapi configurations :-)

observatory-platform/observatory/platform/terraform/main.tf
aroelo: Ah thanks, I saw the added 'google_sql_database', but didn't check the 'google_sql_database_instance', where the network is specified, so I missed this.

tuanchien (approved):

PR #377, opened 2021-03-12
Telescope jstor

aroelo:
Closes https://github.com/The-Academic-Observatory/observatory-platform/issues/246


jdddog (changes requested):
Hey Aniek

Nice work!

Once the Python client API is merged then it would be good to use that to create a telescope and DAG instance for each organisation.

I think we should tweak the paths that files are saved to:
```bash
telescopes/jstor/<organisation name>/upload
telescopes/jstor/<organisation name>/in-progress
telescopes/jstor/<organisation name>/finished
```

We could also tweak the logic for where files are saved after they are downloaded:
* Files deposited to the upload directory.
* When the JSTOR telescope wants to download files via sftp, move them to in-progress directory.
* When the JSTOR telescope has finished (i.e. cleanup) it should move them to the finished directory

It would be good to add a description for each table in BigQuery to make it easy to understand what the purpose of each table is, e.g. for the institutions table: "The number of downloads for open access books authored (?) by UCL press and published on JSTOR, aggregated by month and the institution that downloaded each book".

For the unit tests:
* It would be good to create a fake dataset for testing that can be stored in Github. You could even create it with something such as Faker: https://faker.readthedocs.io/en/master/
* To make testing more flexible, to enable multiple unit test runs to run at once and to make the tests not dependent on the production SFTP server, you could use a Python based SFTP server. An example is https://pypi.org/project/sftpserver/.
  * When the test starts, the Python SFTP server would spin up and the initial files would be deposited.
  * From then on, the SFTP server would be used transparently by the telescope.
  * When the test finishes, it can be shut down.

aroelo:
Thanks for the review Jamie!

> Once the Python client API is merged then it would be good to use that to create a telescope and DAG instance for each organisation.

Yes, definitely.

> I think we should tweak the paths that files are saved to:
> 
> ```shell
> telescopes/jstor/<organisation name>/upload
> telescopes/jstor/<organisation name>/in-progress
> telescopes/jstor/<organisation name>/finished
> ```
> 
> We could also tweak the logic for where files are saved after they are downloaded:
> 
> * Files deposited to the upload directory.
> * When the JSTOR telescope wants to download files via sftp, move them to in-progress directory.
> * When the JSTOR telescope has finished (i.e. cleanup) it should move them to the finished directory

Sounds like a good idea. Might not need it in the end for jstor if we get the reports from gmail, but the same applies for google books, so will make those changes there. I rebased the jstor telescope initially on google books, because they are quite similar. That's also why the google books files were still in there.

> It would be good to add a description for each table in BigQuery to make it easy to understand what the purpose of each table is, e.g. for the institutions table: "The number of downloads for open access books authored (?) by UCL press and published on JSTOR, aggregated by month and the institution that downloaded each book".

I don't think this is possible with the current set-up, because a table is created for each file in the `release.transform_files` list. So there's no way of linking a description for an individual table to a file atm. I could add some functionality to the template so you can give a description if you know the files/tables you expect. 

This would be another property of the Release class
```
table_descriptions = {f'{dag_id}_country': 'A description for the country table',
                      f'{dag_id}_institution': 'A description for the institution table'}
```

And then in any of the bq load methods:
```
        for release in releases:
            for transform_path in release.transform_files:
                transform_blob = blob_name(transform_path)
                table_id, _ = table_ids_from_path(transform_path)
                table_description = release.table_descriptions.get(table_id) <--
```

What do you think?

> For the unit tests:
> 
> * It would be good to create a fake dataset for testing that can be stored in Github. You could even create it with something such as Faker: https://faker.readthedocs.io/en/master/

Thanks, I just created a fake dataset before I saw this comment, but might be useful in the future.

> * To make testing more flexible, to enable multiple unit test runs to run at once and to make the tests not dependent on the production SFTP server, you could use a Python based SFTP server. An example is https://pypi.org/project/sftpserver/.
>   
>   * When the test starts, the Python SFTP server would spin up and the initial files would be deposited.
>   * From then on, the SFTP server would be used transparently by the telescope.
>   * When the test finishes, it can be shut down.

Alright, sounds good, I'll have a look into the python sftp server. If not using it here, then we'll be using it for google books.

jdddog:
> I don't think this is possible with the current set-up, because a table is created for each file in the `release.transform_files` list. So there's no way of linking a description for an individual table to a file atm. I could add some functionality to the template so you can give a description if you know the files/tables you expect.
> 
> This would be another property of the Release class
> 
> ```
> table_descriptions = {f'{dag_id}_country': 'A description for the country table',
>                       f'{dag_id}_institution': 'A description for the institution table'}
> ```
> 
> And then in any of the bq load methods:
> 
> ```
>         for release in releases:
>             for transform_path in release.transform_files:
>                 transform_blob = blob_name(transform_path)
>                 table_id, _ = table_ids_from_path(transform_path)
>                 table_description = release.table_descriptions.get(table_id) <--
> ```
> 
> What do you think?
> 

Yeah that looks good!

observatory-dags/observatory/dags/telescopes/jstor.py
```diff
@@ -0,0 +1,396 @@
...
+            headers = create_headers(report['url'])
+            time.sleep(20)
+            response = requests.get(report['url'], headers=headers)
+            if response.status_code != 200:
+                raise AirflowException(f'Could not get content from download url, reason: {response.reason}, '
+                                       f'status_code: {response.status_code}')
+
+            content = response.content.decode('utf-8')
+            with open(self.download_path(report['type']), 'w') as f:
+                f.write(content)
```
jdddog, line 94: Instead of just waiting for 20 seconds for each download and failing the whole task if one them doesn't download, it would be good to add functionality to retry the download. You can use a library called tenacity, which is already installed with Airflow. We can only use version 4.12.0 however. It would also be good to put variables such as the fixed sleep time into a constant variable. You would need to put the code for downloading the report into it's own function and example is below:

```
from datetime import datetime

from airflow import AirflowException
from tenacity import retry, stop_after_attempt, wait_exponential, wait_fixed

MAX_ATTEMPTS = 3
FIXED_WAIT = 20  # seconds
MAX_WAIT_TIME = 60 * 10  # seconds
EXP_BASE = 3
MULTIPLIER = 10


@retry(
    wait=wait_fixed(FIXED_WAIT)
    + wait_exponential(multiplier=MULTIPLIER, exp_base=EXP_BASE, max=MAX_WAIT_TIME),
    stop=stop_after_attempt(MAX_ATTEMPTS),
    reraise=True,
)
def download_report(status_code):
    print(f"Try {datetime.now().strftime('%H:%M:%S')}")

    if status_code != 200:
        raise AirflowException("API response: {}".format(status_code))
    return "Done!"


for i in range(32):
    download_report(201)
```

observatory-dags/observatory/dags/telescopes/jstor.py
```diff
@@ -0,0 +1,396 @@
...
+    :param service: Gmail service
+    :param publisher: Name of the publisher
+    :param processed_label_id: Id of the 'processed_reports' label
+    :return: Dictionary with release dates as key and reports info as value, where reports info is a list of country
+    and/or institution reports.
+    """
+
+    available_releases = {}
+    # Call the Gmail API
+    results = service.users().messages().list(userId='me', q='subject:"JSTOR Publisher Report Available"').execute()
```
jdddog, line 333:
```suggestion
    results = service.users().messages().list(userId='me', q='subject:"JSTOR Publisher Report Available"', labelIds=["INBOX"]).execute()
```

This is so that the sent items are not included.

observatory-dags/observatory/dags/telescopes/jstor.py
```diff
@@ -0,0 +1,396 @@
...
+            if link.text == 'Download Completed Report':
+                download_url = link['href']
+                break
+        if download_url is None:
+            raise AirflowException(f"Can't find download link for report in e-mail, message snippet: {message.snippet}")
+
+        # get filename from head
+        headers = create_headers(download_url)
+        time.sleep(20)
+        response = requests.head(download_url, headers=headers, allow_redirects=True)
```
jdddog, line 355: I don't think it is necessary to create headers for the Gmail API, it should still work without them?

jdddog (commented):
Hey Aniek, it is looking good. It would be good to use a third party library to make downloading of files more robust (see the comments). Aside from that it would be good for the unit tests to all pass.

observatory-dags/observatory/dags/telescopes/jstor.py
aroelo: Yeah I think you're right. I added these at one point and thought it fixed the bot detection issues, but it seems to work without them as well, so probably a coincidence.

observatory-dags/observatory/dags/telescopes/jstor.py
aroelo: Thanks that was very useful, I've added this. I left the values for the variables the same as in the example, they seem reasonable to me.

observatory-dags/observatory/dags/telescopes/jstor.py
```diff
@@ -0,0 +1,451 @@
...
+    :return: Filename and file extension
+    """
+    logging.info(f'Getting HEAD of report download url, attempt: {get_header_info.retry.statistics["attempt_number"]}, '
+                 f'idle for: {get_header_info.retry.statistics["idle_for"]}')
+    response = requests.head(url, allow_redirects=True)
+    if response.status_code != 200:
+        raise AirflowException(f'Could not get HEAD of report download url, reason: {response.reason}, '
+                               f'status_code: {response.status_code}')
+    filename, extension = response.headers['Content-Disposition'].split('=')[1].split('.')
+    return filename, extension
```
jdddog, line 322: Nice, these look good.

I'm sorry I misunderstood something in the previous review and said that the create_headers function was not needed when calling requests.head, which was wrong. So it would be good to add the following function back and use it to set the headers in requests.get and requests.head in download_report and get_header_info respectively:

```python
def create_headers(url: str) -> dict:
    """ Create a headers dict that can be used to make a request
    :param url: the download url
    :return: headers dictionary
    """
    referer = f"https://www.google.com/url?q={url}" \
              f"&amp;source=gmail&amp;ust={int(time.time())}000&amp;usg=AFQjCNFtACM-4Zqs3yA1AXl4GyEbfvCqwQ"
    headers = {
        'Referer': referer,
        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) '
                      'Chrome/89.0.4389.90 Safari/537.36',
        'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,'
                  'application/signed-exchange;v=b3;q=0.9'

    }
    return headers
```

observatory-dags/observatory/dags/telescopes/jstor.py
```diff
@@ -0,0 +1,451 @@
...
+    The JSTOR telescope.
+
+    Saved to the BigQuery tables: <project_id>.jstor.jstor_countryYYYYMMDD and
+    <project_id>.jstor.jstor_institutionYYYYMMDD
+    """
+
+    DAG_ID_PREFIX = 'jstor'
+    PROCESSED_LABEL_NAME = 'processed_report'
+    ORG_MAPPING = {'anu_press': 'anupress',
+                   'ucl_press': 'uclpress'}
```
jdddog, line 154: Instead of hard coding this, it would be good to use the `extras` field of the telescope e.g. `extras: {"publisher_id": "anupress"}` and then pass the `publisher_id` to the JstorTelescope constructor.

jdddog (commented):
The telescope is looking good, I've just made some small suggestions. It would be good to rebase it to onto the latest version of develop so that the tests will pass.

observatory-dags/observatory/dags/telescopes/jstor.py
aroelo: Ah okay, did you find out that it failed without the headers?

jdddog: Yeah it failed without them

observatory-dags/observatory/dags/telescopes/jstor.py
```diff
@@ -0,0 +1,505 @@
...
+    FIXED_WAIT = 20  # seconds
+    MAX_WAIT_TIME = 60 * 10  # seconds
+    EXP_BASE = 3
+    MULTIPLIER = 10
+
+    def __init__(self, organisation: Organisation, extra: dict, dag_id: Optional[str] = None,
+                 start_date: datetime = datetime(2015, 9, 1), schedule_interval: str = '@monthly',
+                 dataset_id: str = 'jstor', source_format: str = SourceFormat.NEWLINE_DELIMITED_JSON,
+                 dataset_description: str = '', catchup: bool = False, airflow_vars: List = None,
+                 airflow_conns: List = None):
```
jdddog, line 166: I think it would be clearer to specify `publisher_id` as a parameter instead of `extra` in the constructor.

jdddog (changes requested):

observatory-dags/observatory/dags/database/schema/jstor_institution_2021-03-01.json
```diff
@@ -0,0 +1,68 @@
...
+  {
+    "description": "Author of the book.",
+    "mode": "NULLABLE",
+    "name": "Authors",
+    "type": "STRING"
+  },
+  {
+    "description": "ISBN of the book (13 digits).",
+    "mode": "NULLABLE",
+    "name": "ISBN",
```
tuanchien, line 29: Would be helpful if this was type STRING

observatory-dags/observatory/dags/database/schema/jstor_institution_2021-03-01.json
```diff
@@ -0,0 +1,68 @@
...
+    "type": "INTEGER"
+  },
+  {
+    "description": "ISBN of the digital version of the book (13 digits).",
+    "mode": "NULLABLE",
+    "name": "eISBN",
+    "type": "INTEGER"
```
tuanchien, line 36: Would be helpful if this was type STRING

observatory-dags/observatory/dags/database/schema/jstor_institution_2021-03-01.json
aroelo: Thanks for picking that up, I'll check the other mellon telescopes as well to be sure.

jdddog (commented):
It looks good Aniek, lets merge it after rebasing.

jdddog (approved):

PR #380, opened 2021-03-17
Feature Python API Client

jdddog:
This pull request adds the following:
* A Python API client for the Observatory API automatically generated using https://openapi-generator.tech/
* A Github Action for automatically generating an updated Python API client, additional tests and documentation when the OpenAPI specification changes. A pull request is opened with the changes. This can't be tested until this branch is merged.
* Automatically generated documentation for the Python API client and the REST API endpoints, see an [example here](https://observatory-platform--380.org.readthedocs.build/en/380/api/index.html).

The OpenAPI definition and the backend functions have been refactored to work in the most appropriate way for the Python API client.


aroelo (commented):
Hi @jdddog this looks good to me.
I've just had a look through the changed files, but the openapi-generator is new to me so I don't think I can give much feedback. I do have a couple of questions about the usage and will probably look more into the openapi-generator as it seems quite interesting.

What would be the best way to test the updated API client locally if you make changes to the open api config? Instead of waiting for the github workflow. 

What do the host and api key in the configuration refer to?
```
import time
import observatory.api.client
from pprint import pprint
from observatory.api.client.api import observatory_api
from observatory.api.client.model.organisation import Organisation
from observatory.api.client.model.query_response import QueryResponse
from observatory.api.client.model.telescope import Telescope
from observatory.api.client.model.telescope_type import TelescopeType
# Defining the host is optional and defaults to https://api.observatory.academy
# See configuration.py for a list of all supported configuration parameters.
configuration = observatory.api.client.Configuration(
    host = "https://api.observatory.academy"
)

# The client must configure the authentication and authorization parameters
# in accordance with the API server security policy.
# Examples for each auth method are provided below, use the example that
# satisfies your auth use case.

# Configure API key authorization: api_key
configuration.api_key['api_key'] = 'YOUR_API_KEY'
```

Is it possible to use multiple versions in the python API client? I saw that there is the 'queryv1' method (https://observatory-platform--380.org.readthedocs.build/en/380/api/ObservatoryApi.html#queryv1), I assume the name is based on the function name. Right now all methods are under the '/v1' path. Would it be possible to create an 'api_instance' for v1 or for v2, so you can set the version for the entire instance, rather than picking a version for an individual method?

jdddog:
> Hi @jdddog this looks good to me.
> I've just had a look through the changed files, but the openapi-generator is new to me so I don't think I can give much feedback. I do have a couple of questions about the usage and will probably look more into the openapi-generator as it seems quite interesting.

Thanks.

> 
> What would be the best way to test the updated API client locally if you make changes to the open api config? Instead of waiting for the github workflow.

You could run the commands in the Github Actions workflow on your machine to generate a new Python API client.

> 
> What do the host and api key in the configuration refer to?
> 
> ```
> import time
> import observatory.api.client
> from pprint import pprint
> from observatory.api.client.api import observatory_api
> from observatory.api.client.model.organisation import Organisation
> from observatory.api.client.model.query_response import QueryResponse
> from observatory.api.client.model.telescope import Telescope
> from observatory.api.client.model.telescope_type import TelescopeType
> # Defining the host is optional and defaults to https://api.observatory.academy
> # See configuration.py for a list of all supported configuration parameters.
> configuration = observatory.api.client.Configuration(
>     host = "https://api.observatory.academy"
> )
> 
> # The client must configure the authentication and authorization parameters
> # in accordance with the API server security policy.
> # Examples for each auth method are provided below, use the example that
> # satisfies your auth use case.
> 
> # Configure API key authorization: api_key
> configuration.api_key['api_key'] = 'YOUR_API_KEY'
> ```
> 

`host` is the URL of the API, you can change this so that the client library interacts with different versions of the API hosted on different domains, e.g. `develop.api.observatory.academy` or `aniek-dev.api.observatory.academy`. 

`api_key` is the API key for authorisation, i.e. the API key that you generate for Cloud Endpoints. 

> Is it possible to use multiple versions in the python API client? I saw that there is the 'queryv1' method (https://observatory-platform--380.org.readthedocs.build/en/380/api/ObservatoryApi.html#queryv1), I assume the name is based on the function name. Right now all methods are under the '/v1' path. Would it be possible to create an 'api_instance' for v1 or for v2, so you can set the version for the entire instance, rather than picking a version for an individual method?

I'm not sure, we could look at that as future work. It would be good if you could do that.

aroelo (approved):

PR #386, opened 2021-03-31
Feature google analytics

aroelo:


jdddog (changes requested):
Hey Aniek, it looks good. It would be good to use the extras field from the API to store the view_id and the page_path_regex and also rebase onto the latest develop branch.

observatory-dags/observatory/dags/telescopes/google_analytics.py
```diff
@@ -0,0 +1,515 @@
...
+        else:
+            return False
+
+
+class GoogleAnalyticsTelescope(SnapshotTelescope):
+    """ Google Analytics Telescope."""
+    DAG_ID_PREFIX = 'google_analytics'
+
+    def __init__(self, organisation: Organisation, extra: dict, dag_id: Optional[str] = None,
+                 start_date: datetime = datetime(2021, 1, 1),
```
jdddog, line 105: I would prefer to have view_id and pagepath_regex as explicit parameters in the constructor (rather than passing the extra dictionary), because it makes the GoogleAnalyticsTelescope constructor interface clearer.

jdddog (changes requested):

jdddog (commented):
It looks good Aniek, lets merge it after rebasing.

jdddog (approved):

PR #388, opened 2021-03-31
Add ONIX telescope

jdddog:
This pull request adds the ONIX telescope. A DAG is created for each ONIX telescope that has been added via the Observatory REST API. Some of the code, e.g. bq_load, could be tidied up a bit and merged into the SnapshotTelscope, but I think this would be better done in another pull request, as we might need to change other telescopes as well.

JDK 11 has been added to the following so that ONIX parser can run (which is a java jar):
* Github Actions unit-tests.yml
* Dockerfile

The Docker related files and the local and Terraform platform builders have been modified to include the observatory-api package in the Docker builds.

### DIFF
--- a/observatory-dags/observatory/dags/telescopes/onix.py
+++ b/observatory-dags/observatory/dags/telescopes/onix.py
@@ -223,18 +223,17 @@ def list_release_info(sftp_upload_folder: str) -> List[Dict]:
     """
 
     results = []
-    sftp = make_sftp_connection()
-    sftp.makedirs(sftp_upload_folder)
-    files = sftp.listdir(sftp_upload_folder)
-    for file_name in files:
-        if re.match(OnixRelease.DOWNLOAD_FILES_REGEX, file_name):
-            date_str = file_name[:8]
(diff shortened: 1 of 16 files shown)


observatory-dags/observatory/dags/telescopes/onix.py
```diff
@@ -0,0 +1,382 @@
...
+        :return: None.
+        """
+
+        with make_sftp_connection() as sftp:
+            sftp.makedirs(self.sftp_finished_folder)
+            sftp.rename(self.sftp_in_progress_file, self.sftp_finished_file)
+
+
+def org_id(organisation_name: str):
+    return organisation_name.strip().replace(' ', '_').lower()
```
jdddog, line 205: Docstring

aroelo (approved):
Hey Jamie,

Looks good!
As discussed on slack we can hopefully make some changes in the next PR related to the template working with different projects.

jdddog:
> Hey Jamie,
> 
> Looks good!
> As discussed on slack we can hopefully make some changes in the next PR related to the template working with different projects.

Thanks Aniek, yeah that sounds good.

PR #392, opened 2021-04-12
Add sftp utils

aroelo:
Add SftpFolders class, so sftp functions can be used from multiple telescopes.
Implemented usage of SftpFolders for the Onix telescope.

### DIFF
--- a/observatory-dags/observatory/dags/telescopes/onix.py
+++ b/observatory-dags/observatory/dags/telescopes/onix.py
@@ -94,7 +94,6 @@ class OnixRelease(SnapshotRelease):
     def download(self):
         """ Downloads an individual ONIX release from the SFTP server.
 
-        :param in_progress_folder: Sftp 'in_progress' folder
         :return: None.
         """
 
--- a/observatory-dags/observatory/dags/telescopes/onix.py
+++ b/observatory-dags/observatory/dags/telescopes/onix.py
@@ -57,6 +57,7 @@ class OnixRelease(SnapshotRelease):
         super().__init__(dag_id, release_date, self.DOWNLOAD_FILES_REGEX, None, self.TRANSFORM_FILES_REGEX)
         self.organisation = organisation
         self.file_name = file_name
+        self.sftp_folders = SftpFolders(dag_id, organisation.name)
 
     @property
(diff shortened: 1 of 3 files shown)


observatory-dags/observatory/dags/telescopes/onix.py
```diff
@@ -138,24 +83,15 @@ def transform_bucket(self):
         """
         return self.organisation.gcp_transform_bucket
 
-    def move_files_to_in_progress(self):
```
jdddog, line 141: I would prefer to leave some of these functions in here, e.g. `move_files_to_finished`, `download` (with the interface that it had) and `move_files_to_finished` (these are the interface that the telescope uses) and use the SftpFolder class from within the OnixRelease object.

jdddog (commented):
That looks good Aniek. The thing that I would like to change is to treat the SftpFolders object as an implementation detail of the OnixRelease class, so to instantiate a new SftpFolders object within the OnixRelease constructor and use it within each method within the OnixRelease object to get the desired result. I prefer to keep the implementation for processing a release within the release object.

aroelo:
> That looks good Aniek. The thing that I would like to change is to treat the SftpFolders object as an implementation detail of the OnixRelease class, so to instantiate a new SftpFolders object within the OnixRelease constructor and use it within each method within the OnixRelease object to get the desired result. I prefer to keep the implementation for processing a release within the release object if possible.

Yeah I was doubting myself whether to put it in the OnixRelease or in the OnixTelescope class.
I decided to go with the OnixTelescope, because to me it makes sense to try to only have release specific properties in the release class (e.g. things dependent on the release date). But what you say about the implementation makes sense as well.

The upload folder is already required before making a release in the 'list_release_info' function, so I did have to add the SftpFolders there as well.

I just updated the Onix telescope, what do you think of the latest version?

observatory-dags/observatory/dags/telescopes/onix.py
aroelo: This should be resolved in the commit I just made, unless I misunderstood.

jdddog (approved):
Looks good to me

PR #394, opened 2021-04-14
Feature/add sensors telescope

tuanchien:
- Introduce "add_sensors" to telescope interface that allows telescopes to block tasks until sensors are ready.
- Introduce a generic WorkflowTelescope for workflow telescope use.


aroelo (commented):
Hey Tuan,

Do you already have a couple of 'workflow' telescopes?
Just asking, because I'm curious if there are likely more distinctive properties/methods than the sensor and the fact that it does not use a release.

It is already possible to add a sensor as a task using the 'Telescope' class.
Just a simple sensor as an example:
```python
import logging
from datetime import datetime

from airflow.sensors.base_sensor_operator import BaseSensorOperator


class MySensor(BaseSensorOperator):
    def __init__(self, *args, **kwargs):
        task_id = 'my_sensor'
        super(MySensor, self).__init__(task_id=task_id, *args, **kwargs)

    def poke(self, context):
        current_minute = datetime.now().minute
        if current_minute % 3 != 0:
            logging.info("Current minute (%s) not is divisible by 3, sensor will retry.", current_minute)
            return False

        logging.info("Current minute (%s) is divisible by 3, sensor finishing.", current_minute)
        return True
```

A telescope with sensors:

```python
from datetime import datetime
from functools import partial

from observatory.platform.telescopes.telescope import Telescope


class MyWorkflowTelescope(Telescope):
    def __init__(self, dag_id: str = 'workflow', start_date: datetime = datetime(2018, 5, 14),
                 schedule_interval: str = '@weekly'):
        """ Construct a Workflow instance.

        :param dag_id: the id of the DAG.
        :param start_date: the start date of the DAG.
        :param schedule_interval: the schedule interval of the DAG.
        """

        super().__init__(dag_id, start_date, schedule_interval)

        self.add_setup_task(partial(MySensor, queue=self.queue, default_args=self.default_args))
        self.add_setup_task(self.download)

    def download(self, **kwargs):
        print('hello')

    def make_release(self) -> None:
        """ Not needed. """
        return None
```

So it might not be necessary to add the 'add_sensors' method to the telescope.
If the sensors are often used in 'workflow' telescope types, perhaps it would be useful to create the WorkflowTelescope class and add some helper methods as well as implement the `make_release` method like you did.

That could look something like this:
```python
from datetime import datetime
from functools import partial
from typing import Callable, List, Type

from airflow.sensors.base_sensor_operator import BaseSensorOperator
from observatory.platform.telescopes.telescope import Telescope


class WorkflowTelescope(Telescope):
    def __init__(self, dag_id: str, start_date: datetime, schedule_interval: str, catchup: bool = True,
                 queue: str = 'default', max_retries: int = 3, max_active_runs: int = 1,
                 airflow_vars: list = None, airflow_conns: list = None):
        """ Construct a WorkflowTelescope instance.

        :param dag_id: the id of the DAG.
        :param start_date: the start date of the DAG.
        :param schedule_interval: the schedule interval of the DAG.
        :param catchup: whether to catchup the DAG or not.
        :param queue: the Airflow queue name.
        :param max_retries: the number of times to retry each task.
        :param max_active_runs: the maximum number of DAG runs that can be run at once.
        :param airflow_vars: list of airflow variable keys, for each variable it is checked if it exists in airflow
        :param airflow_conns: list of airflow connection keys, for each connection it is checked if it exists in airflow
        """
        super().__init__(dag_id, start_date, schedule_interval, catchup, queue, max_retries, max_active_runs,
                         airflow_vars, airflow_conns)

    def make_sensor(self, sensor: Type[BaseSensorOperator]) -> Callable:
        """ Make a partial BaseSensorOperator that can be attached to an airflow DAG.
        :param sensor: A BaseSensorOperator or child of BaseSensorOperator
        :return: The partial BaseSensorOperator
        """
        operator = partial(sensor, queue=self.queue, default_args=self.default_args, )

        return operator

    def make_sensors(self, sensors: List[Type[BaseSensorOperator]]) -> List[Callable]:
        """ Make a partial BaseSensorOperator for each sensor in the list.
        :param sensors: A list of BaseSensorOperators
        :return: A list of partial BaseSensorOperator
        """
        operators = []
        for sensor in sensors:
            operator = self.make_sensor(sensor)
            operators.append(operator)
        return operators

    def make_release(self) -> None:
        """ Not needed. """
        return None
```

Then the telescope would be a little bit simpler:
```python
from datetime import datetime
from observatory.platform.telescopes.workflow_telescope import WorkflowTelescope


class MyWorkflowTelescope(WorkflowTelescope):
    def __init__(self, dag_id: str = 'workflow', start_date: datetime = datetime(2018, 5, 14),
                 schedule_interval: str = '@weekly'):
        """ Construct a Workflow instance.

        :param dag_id: the id of the DAG.
        :param start_date: the start date of the DAG.
        :param schedule_interval: the schedule interval of the DAG.
        """

        super().__init__(dag_id, start_date, schedule_interval)

        self.add_setup_task(self.make_sensor(MySensor))
        self.add_setup_task(self.download)

    def download(self, **kwargs):
        print('hello')

    def make_release(self) -> None:
        """ Not needed. """
        return None
```

I have set-up something similar for the StreamTelescope, because some tasks use a slightly different PythonOperator.

tuanchien:
Thanks for the comments Aniek.

The workflow I have in mind needs to wait for a telescope to finish (ExternalTaskSensor) before executing a workflow.  The only other workflow I know of (doi one) is a similar setup I believe.  So from this point of view it is really just sensors and no release. Introducing a separate workflow telescope was just for conceptual organisation. I'm happy to just go off the base Telescope class if that's what we all want to do.

Similarly the extra sensor methods are mostly for conceptual separation. As you pointed out, they could be put in the workflow telescope instead. That does mean sensors always need to be loaded on before other setup tasks though since they are just being appended to the setup list.  If we just want to re-use setup_tasks for sensors, I'm happy to do this too.

observatory-platform/observatory/platform/telescopes/telescope.py
```diff
@@ -88,6 +89,16 @@ def add_setup_task_chain(self, funcs: List[Callable]):
...
+    @abstractmethod
+    def add_sensors(self, sensors: List[BaseSensorOperator]):
+        """ Add a list of sensors to monitor.  The telescope will wait until the monitored sensors all trigger before
+        running the tasks.
+
+        :param sensors: List of sensors to monitor.
+        """
+        pass
+
+
```
jdddog, line 101:
```suggestion
```

observatory-platform/observatory/platform/telescopes/telescope.py
```diff
@@ -88,6 +89,16 @@ def add_setup_task_chain(self, funcs: List[Callable]):
         """
         pass
 
+    @abstractmethod
+    def add_sensors(self, sensors: List[BaseSensorOperator]):
```
jdddog, line 93: It would be good to also add an interface `add_sensor` so that you can add a single sensor at a time if you wish.

observatory-platform/observatory/platform/telescopes/telescope.py
```diff
@@ -272,6 +292,9 @@ def make_dag(self) -> DAG:
                 tasks.append(task)
             chain(*tasks)
 
+            # Add sensors as a pre-requisite to the first task.
+            self.sensors >> tasks[0]
```
jdddog, line 296: What happens if self.sensors is an empty list?

observatory-platform/observatory/platform/telescopes/workflow_telescope.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Generic Workflow telescope for running tasks.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def make_release(self) -> None:
+        """ Not needed. """
+        return None
```
jdddog, line 31: I don't know if it is necessary to make a new class? You might still want to create a Release object when running a workflow, e.g. to store a date or version for the processed dataset.

jdddog (commented):
Thanks Tuan, that looks good, I have left a few small comments.

jdddog:
I agree that the extra WorkflowTelescope class is unnecessary. 

> Do you already have a couple of 'workflow' telescopes?
> Just asking, because I'm curious if there are likely more distinctive properties/methods than the sensor and the fact that it does not use a release.

You might want to use a release for a workflow, e.g. to store the date of the release for the dataset that is created by the workflow,  some other identifier or perhaps the versions of each dataset that is harvested.

> It is already possible to add a sensor as a task using the 'Telescope' class.
> Just a simple sensor as an example:

I think the interface for adding sensors is clearer by using `add_sensor` and `add_sensors` as methods.

Also, we want to be able to wait for multiple sensors in parallel before the tasks start, e.g. like the DOI workflow below (adding tasks in parallel is out of scope for this pull request), so make_dag will need to be modified to handle that.

image

tuanchien:
Taking into account your POVs, how about:
 - getting rid of WorkflowTelescope and just using Telescope base class for workflows
 - adding add_sensor/add_sensors to Telescope base class
 - implementing a helper function for creating ExternalTaskSensor from telescope Responses (observatory api call response) in a util module that workflow creators can choose to use if they wish.

aroelo:
> Taking into account your POVs, how about:
> 
> * getting rid of WorkflowTelescope and just using Telescope base class for workflows,
> * adding add_sensor/add_sensors to Telescope base class,
> * implementing a helper function for creating ExternalTaskSensor from telescope Responses (observatory api call response) in a util module that workflow creators can choose to use if they wish,
> * letting each workflow handle its own releases.

Yeah that sounds good to me!

jdddog (approved):

PR #395, opened 2021-04-15
Books extension to doi workflow

rhosking:


rhosking:
@jdddog any suggestions on additional testing? It's essentially an additionally SQL query, but it does raise the question again around how we go about unit testing those

PR #396, opened 2021-04-16
Extend telescope template to add table descriptions per telescope

aroelo:


jdddog (approved):
That looks good to me Aniek

PR #401, opened 2021-04-21
Bugfix/find schema

tuanchien:
- Slightly simplified the schema file naming spec to: [prefix]tablename_[ver_]YYYY-MM-DD.json  where [] indicates optional bits.
- Renamed the schema files to be compliant with new spec, i.e., adding _ to the Mag ones between tablename and date.
- Handle the case where one table name is a prefix of another table name, e.g., tableone, tableonenew.


jdddog (changes requested):
Looks good to me Tuan, it just needs to be rebased

jdddog (approved):

PR #402, opened 2021-04-21
Feature extending doi workflow

rhosking:
- added additional MAG fields to the DOIs table
- additional fields from crossref added to the DOIs table
- cleaned up double nesting issues within the DOIs table for various MAG fields
- updated aggregate_doi script to reflect the field changes
- added clustering support to load_bigquery_table method
- improved the crossref_events aggregration step
- added new create_book extension to the DOIs workflow
- added new relationship tracking to the aggregation tables which replaces collaborations (which often made no sense)
- added percentage of total to relationships


jdddog (approved):
I will be adding unit test for the DOIs DAG in another pull request.

PR #403, opened 2021-04-22
Feature/onix workflow

tuanchien:
Workflow telescope that waits on ONIX telescope, aggregates into works and work families, and produces bigquery lookup tables for them.

- oaebu intermediate tables will be handled in a separate branch
- there's been no testing of externaltasksensor code. that's probably better left for integration testing.

### DIFF
--- a/observatory-dags/observatory/dags/workflows/onix_workflow.py
+++ b/observatory-dags/observatory/dags/workflows/onix_workflow.py
@@ -19,6 +19,7 @@ import json
 import os
 import shutil
 from datetime import datetime
+from functools import partial, update_wrapper
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
@@ -27,7 +28,9 @@ import pendulum
 from airflow.sensors.external_task_sensor import ExternalTaskSensor
 from google.cloud.bigquery import SourceFormat
 from jinja2 import Environment, PackageLoader
+from observatory.dags.config import workflow_sql_templates_path
 from observatory.dags.telescopes.onix import OnixTelescope
+from observatory.dags.workflows.oaebu_partners import OaebuPartners
 from observatory.dags.workflows.onix_work_aggregation import (
     BookWorkAggregator,
     BookWorkFamilyAggregator,
@@ -35,10 +38,14 @@ from observatory.dags.workflows.onix_work_aggregation import (
 from observatory.platform.telescopes.telescope import AbstractRelease, Telescope
 from observatory.platform.utils.airflow_utils import AirflowConns, AirflowVars
 from observatory.platform.utils.gc_utils import (
+    bigquery_partitioned_table_id,
+    create_bigquery_dataset,
+    create_bigquery_table_from_query,
     run_bigquery_query,
     select_table_suffixes,
     upload_files_to_cloud_storage,
 )
+from observatory.platform.utils.jinja2_utils import render_template
 from observatory.platform.utils.telescope_utils import (
     list_to_jsonl_gz,
     make_dag_id,
@@ -100,6 +107,9 @@ class OnixWorkflowRelease(AbstractRelease):
         self.bucket_name = gcp_bucket_name
 
         Path(".", self.transform_folder).mkdir(exist_ok=True, parents=True)
+
+        # OAEBU intermediate tables
+        self.oaebu_intermediate_dataset = "oaebu_intermediate"
 
     @property
     def transform_bucket(self) -> str:
@@ -177,6 +187,9 @@ class OnixWorkflow(Telescope):
     3. Create an ISBN13 -> Work Family ID lookup table.  Clusters editions together.
       a. Aggregate Works into Work Families.
       b. Writes the lookup table to BigQuery.
+    4. Create OAEBU intermediate tables.
+      a. For each data partner, create new tables in oaebu_intermediate dataset where existing tables are augmented
+         with work_id and work_family_id columns.
     """
 
     DAG_ID_PREFIX = "onix_workflow"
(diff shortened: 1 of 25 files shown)


observatory-dags/observatory/dags/workflows/onix_workflow.py
```diff
@@ -0,0 +1,314 @@
...
+        self.local_table_dir = self.blob_path
+        self.workslookup_filename = os.path.join(self.local_table_dir, "onix_workid_isbn.jsonl.gz")
+        self.workslookup_errors_filename = os.path.join(self.local_table_dir, "onix_workid_isbn_errors.jsonl.gz")
+        self.worksfamilylookup_filename = os.path.join(self.local_table_dir, "onix_workfamilyid_isbn.jsonl.gz")
+
+        Path(".", self.local_table_dir).mkdir(exist_ok=True, parents=True)
+
+        # Work (family) aggregation results
+        self.works = list()
+        self.works_families = list()
```
jdddog, line 122: It would be good to put these into a release object similar to the other DAGs and create the release object as part of an Airflow task.

The reason is that, the OnixWorkflow is constructed, and therefore the `release_date` is fetched, when the DAG file is loaded, not after waiting for the ONIX telescope to finish (via the external task sensor).

observatory-dags/observatory/dags/workflows/onix_workflow.py
```diff
@@ -0,0 +1,314 @@
...
+        """Get the newest table name from the onix dataset.
+
+        :return: Latest table name and date string.
+        """
+
+        sql = f"SELECT table_name from {self.project_id}.{self.onix_dataset_id}.INFORMATION_SCHEMA.TABLES ORDER BY table_name DESC LIMIT 1"
+        df = self.client.query(sql).to_dataframe()
+        table_name = df["table_name"][0]
+        date_str = table_name[-8:]  # Last 8 digits is the YYYYMMDD date string
+        return table_name, date_str
```
jdddog, line 164: There is already a function here that can be used to select the table shard dates for a particular table: https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory-dags/observatory/dags/workflows/doi.py#L50-L68

It could be moved into the telescope utilities and renamed into something that makes better sense. You can also specify an end date to search up to (so that you only include tables up to the release date of the workflow for instance, in the case where there might be future releases of a table, such as when back-filling).

I think you can assume that the ONIX tables will always be called `onix`, so that could just be hardcoded as a constant in the telescope.

jdddog (changes requested):
Hey Tuan, it looks good, just a couple of comments.

tuanchien:
- Refactored to use select_table_suffix, and put that function in gc_utils
- Using a release object now.  Not sure if this is particularly fit for purpose though since half of the methods in AbstractRelease are unused.
- Do all the in-memory work(family) aggregation in a single task to avoid needing to share info between tasks.
- Changed ISBN/GTIN in onix schema to string instead of integer.
- Added tasks for creating oaebu_intermediate tables that have work(family) ids added. This was initially intended for an independent branch.  There is a temporary hard coded method for populating the data partner info until we get a more long term solution in place later.

observatory-dags/observatory/dags/workflows/onix_workflow.py
```diff
@@ -0,0 +1,509 @@
...
+
+        release_date = kwargs["execution_date"]
+        release = OnixWorkflowRelease(
+            dag_id=self.dag_id,
+            release_date=release_date,
+            gcp_project_id=self.gcp_project_id,
+            gcp_bucket_name=self.gcp_bucket_name,
+            onix_dataset_id=self.onix_dataset_id,
+            onix_table_id=self.onix_table_id,
+        )
```
jdddog, line 271: The biggest thing that will need changing is refactoring the OnixWorkflow to process multiple releases. The OnixTelescope processes multiple releases for each DAG run (e.g. if multiple ONIX files have been deposited), therefore the OnixWorkflow should also be able to process multiple releases.

I'm thinking that a better way to communicate what releases need to be processed by the OnixWorkflow, is to pull the XCom published by the OnixTelescope for the same period. This XCom contains a list of release dates, e.g. `[{'release_date': <Pendulum [2021-04-20T00:00:00+00:00]>, 'file_name': '20210420_ANUPRESS_ONIX.xml'}]`. This means that you would not use `select_table_suffixes` further down, just the release dates pulled from XCom stored in the release object.

You can pull them like so:
```python
ti: TaskInstance = kwargs['ti']
records = ti.xcom_pull(
    dag_id='onix_telescope_name',
    key=OnixTelescope.RELEASE_INFO,
    task_ids='list_release_info',
    include_prior_dates=False)
```

jdddog (changes requested):
It is looking good Tuan, see the comments below for suggestions for changes.

observatory-dags/observatory/dags/workflows/onix_workflow.py
```diff
@@ -0,0 +1,509 @@
...
+        self.onix_dataset_id = onix_dataset_id
+        self.onix_table_id = onix_table_id
+
+        # Initialise Telesecope base class
+        super().__init__(
+            dag_id=self.dag_id, start_date=start_date, schedule_interval=schedule_interval, catchup=catchup
+        )
+
+        # Wait on ONIX telescope to finish
+        self.add_sensor(telescope_sensor)
```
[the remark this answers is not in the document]
tuanchien: I was trying to avoid having a dependency on Observatory API classes by passing in a sensor pre-constructed. Would you prefer the org name gets passed to the telescope?

jdddog (changes requested):
Hey Tuan, that looks good, just a couple of minor suggestions and then I think we can merge it.

jdddog (approved):

PR #404, opened 2021-04-22
Update telescope template 'make_dag'

aroelo:
Add functionality to check whether a function part of 'setup_task_funcs' or 'task_funcs' is a partial operator.
Previously this only checked whether it is a partial function.

This makes it possible to pass on a partial function to the PythonOperator, instead of assuming that the partial function is an airflow operator.

### DIFF
--- a/observatory-dags/observatory/dags/telescopes/google_analytics.py
+++ b/observatory-dags/observatory/dags/telescopes/google_analytics.py
@@ -21,13 +21,11 @@ import datetime
 import logging
 import os
 from datetime import datetime
-from functools import partial
 from typing import Dict, List, Optional, Tuple
 
 import pendulum
-from airflow.exceptions import AirflowException
+from airflow.exceptions import AirflowException, AirflowSkipException
 from airflow.hooks.base_hook import BaseHook
-from airflow.operators.python_operator import ShortCircuitOperator
 from googleapiclient.discovery import Resource, build
 from oauth2client.service_account import ServiceAccountCredentials
 from observatory.api.client.model.organisation import Organisation
@@ -137,11 +135,8 @@ class GoogleAnalyticsTelescope(SnapshotTelescope):
         self.pagepath_regex = pagepath_regex
 
         self.add_setup_task_chain([self.check_dependencies])
-        self.add_task(partial(ShortCircuitOperator, task_id=self.download_transform.__name__,
-                              python_callable=partial(self.task_callable, self.download_transform),
-                              queue=self.queue, default_args=self.default_args,
-                              provide_context=True))
-        self.add_task_chain([self.upload_transformed,
+        self.add_task_chain([self.download_transform,
+                             self.upload_transformed,
                              self.bq_load,
                              self.cleanup])
 
@@ -181,8 +176,9 @@ class GoogleAnalyticsTelescope(SnapshotTelescope):
         :param releases: a list with one google analyics release.
         :return: None.
         """
-        success = releases[0].download_transform(self.view_id, self.pagepath_regex)
-        return success
+        results = releases[0].download_transform(self.view_id, self.pagepath_regex)
+        if not results:
+            raise AirflowSkipException("No Google Analytics data available to download.")
 
     def bq_load(self, releases: List[SnapshotRelease], **kwargs):
         """ Task to load each transformed release to BigQuery.
--- a/observatory-dags/observatory/dags/telescopes/oapen_irus_uk.py
+++ b/observatory-dags/observatory/dags/telescopes/oapen_irus_uk.py
@@ -229,10 +229,7 @@ class OapenIrusUkTelescope(SnapshotTelescope):
 
         self.add_setup_task(self.check_dependencies)
         # create PythonOperator with task concurrency of 1, so tasks to create cloud function never run in parallel
-        self.add_task(partial(PythonOperator, task_id=self.create_cloud_function.__name__,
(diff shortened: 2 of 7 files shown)


tuanchien:
Looks good Aniek.  I had a thought this morning though about maybe a way to unify it. Let me know if this is viable.

So it seems like make_operators https://github.com/The-Academic-Observatory/observatory-platform/blob/f08088ee1a82c2cffb29fb1ae99943e6ece740a7/observatory-platform/observatory/platform/telescopes/stream_telescope.py#L105

differs only in the trigger rule.

What if we just made add_task(_chain) pass the kwargs onto the PythonOperator constructor instead?

So add_task would look like
`def add_task(self, func: Callable, **kwargs)`

similarly for `add_task_chain`

and make_dag instead of branching can do things like
```
task = ShortCircuitOperator(
  task_id=func.__name__,
  python_callable=func,
  queue=self.queue,
  default_args=self.default_args,
  provide_context=True,
  **kwargs
)

task = PythonOperator(
  task_id=func.__name__,
  python_callable=partial(self.task_callable, func),
  queue=self.queue,
  default_args=self.default_args,
  provide_context=True,
  **kwargs
)
```

Then instead of having to call `make_operator` or `make_operators` in StreamTelescope, you can just use
`add_task(yourtask, trigger_rule="none_failed")` or `add_task_chain(..., trigger_rule="none_failed")`.

aroelo:
> Looks good Aniek. I did have a thought this morning though about maybe a way to unify it. Let me know if this is viable, otherwise we'll go with what you currently have.
> 
> So it seems like make_operators
> 
> https://github.com/The-Academic-Observatory/observatory-platform/blob/f08088ee1a82c2cffb29fb1ae99943e6ece740a7/observatory-platform/observatory/platform/telescopes/stream_telescope.py#L105
> 
> differs only in the trigger rule.
> 
> What if we just made make_dag pass the kwargs onto the PythonOperator constructor instead?
> 
> So make_dag in base Telescope would look like
> `def make_dag(self, **kwargs)`
> 
> and make_dag instead of branching can do things like
> 
> ```
> task = PythonOperator(
>   task_id=func.__name__,
>   python_callable=partial(self.task_callable, func),
>   queue=self.queue,
>   default_args=self.default_args,
>   provide_context=True,
>   **kwargs
> )
> ```
> 
> The user would either pass the trigger rule invoking make_dag(), or StreamTelescope can override make_dag to do something like:
> 
> ```
>   def make_dag(self):
>     return super().make_dag(trigger_rule='none_failed')
> ```
> 
> An alternative is to modify add_task(_chain) instead, to track the kwargs, but that's a much bigger undertaking.

Mmmm, yes for the StreamTelescope the only difference is the trigger_rule, it would be a bit easier to just pass the kwargs to the make_dag, but for the StreamTelescope the trigger_rule should only be different for some tasks, not all tasks of the DAG.

Specifically, the trigger rule for some bigquery loading tasks is set to 'none_failed', so upstream tasks can be skipped, and these tasks will still run. The default behaviour is set to 'all_succeed', meaning that if any upstream tasks are skipped then all downstream tasks are skipped as well.

Maybe we could modify add_task to track the kwargs. I had a think about this when developing the template, and liked the possibility to add a complete operator, so you could pass an operator different than the PythonOperator.

tuanchien:
Ok, how about:
```
def add_task(self, func: Callable, **kwargs):
  self.task_funcs.append((func, kwargs))

def add_task_chain(self, funcs: List[Callable], **kwargs):
  self.task_funcs += [(func, kwargs) for func in funcs]
```

and then in make_dags:
```
...
for func, kwargs in self.task_funcs:
  task = PythonOperator(
                        task_id=func.__name__,
                        python_callable=partial(self.task_callable, func),
                        queue=self.queue,
                        default_args=self.default_args,
                        provide_context=True,
                        **kwargs,
                    )
  tasks.append(task)
...
```

jdddog:
> Ok, how about:
> 
> ```
> def add_task(self, func: Callable, **kwargs):
>   self.task_funcs.append((func, kwargs))
> 
> def add_task_chain(self, funcs: List[Callable], **kwargs):
>   self.task_funcs += [(func, kwargs) for func in funcs]
> ```
> 
> and then in make_dags:
> 
> ```
> ...
> for func, kwargs in self.task_funcs:
>   task = PythonOperator(
>                         task_id=func.__name__,
>                         python_callable=partial(self.task_callable, func),
>                         queue=self.queue,
>                         default_args=self.default_args,
>                         provide_context=True,
>                         **kwargs,
>                     )
>   tasks.append(task)
> ...
> ```

That looks good to me.

aroelo:
> Ok, how about:
> 
> ```
> def add_task(self, func: Callable, **kwargs):
>   self.task_funcs.append((func, kwargs))
> 
> def add_task_chain(self, funcs: List[Callable], **kwargs):
>   self.task_funcs += [(func, kwargs) for func in funcs]
> ```
> 
> and then in make_dags:
> 
> ```
> ...
> for func, kwargs in self.task_funcs:
>   task = PythonOperator(
>                         task_id=func.__name__,
>                         python_callable=partial(self.task_callable, func),
>                         queue=self.queue,
>                         default_args=self.default_args,
>                         provide_context=True,
>                         **kwargs,
>                     )
>   tasks.append(task)
> ...
> ```

Sounds good, I'll make some changes so they can be reviewed again :-)

tuanchien (approved):

PR #381, opened 2021-03-18
Telescope doab

aroelo:
Closing https://github.com/The-Academic-Observatory/observatory-platform/issues/304

PR #405, opened 2021-04-23
Add functionality to use dag run in test environment.

aroelo:
This PR adds the functionality to use a dag_run in the test environment, this is useful because:
- DAG runs might be different for a first run compared to a second run
For example for the 'StreamTelescope', there are the following tasks for loading data into BigQuery:

1. bq_load_partition. Load data into a single BigQuery partition
2. bq_delete_old. Delete old rows from the main table.
3. bq_append_new. Append new rows to the main table.

These tasks are slightly different for a first run, compared to a second (or later) run.
Task 1 is skipped for the first run, so the data is just appended to the main table in task 3. The partition and the main table would be the same. Since the first run most likely contains all data from the past up until now this can be big in size, it saves some resources to just add it to the main table and not as a partition as well.
For a later run this is where the data from that current release/file is loaded into a partition.

Task 2 is also skipped for the first run, because there is no old data.
For a later run the data from the relevant partition(s) is deleted from the main table. It can be 1 partition or multiple partitions, depending on how many partitions have been added since this task last ran (defined by bq_merge_days).

Task 3 is executed for both first and later runs, but for the first run the data is added to the table from a file, while for the later run the data is added from the partition that is created in task 1.
For a later run the data from the relevant partition(s) is added to the main table. It can be 1 partition or multiple partitions, depending on how many partitions have been added since this task last ran (defined by bq_merge_days).

- With the addition of the dag_run, tasks can be assigned to a specific dag_run and it is possible to check the state of a previous Task Instance for a certain task.
- The start date of the DAG Run can be set to a chosen point in time. This is useful, because some tasks use the number of days between the current run and the previous successful run.

Most changes were originally made in the DOAB telescope pull request (https://github.com/The-Academic-Observatory/observatory-platform/pull/381)

### DIFF
--- a/observatory-platform/observatory/platform/utils/test_utils.py
+++ b/observatory-platform/observatory/platform/utils/test_utils.py
@@ -70,6 +70,8 @@ import uuid
 from functools import partial
 from typing import Dict
 
+import croniter
+import datetime
 import httpretty
 import paramiko
 import pendulum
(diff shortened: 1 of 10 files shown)


observatory-platform/observatory/platform/utils/test_utils.py
```diff
@@ -249,26 +257,61 @@ def add_connection(self, conn: Connection):
...
         :param dag: the Airflow DAG instance.
         :param task_id: the Airflow task identifier.
         :param execution_date: the execution date of the DAG.
         :return: None.
         """
-
+        # If dag or execution date are not set, get corresponding values from dag run
+        if not (dag and execution_date):
+            if self.dag_run is None:
+                raise Exception(
```
aroelo, line 271: Not sure what the best way to exit / raise an exception is and make the test fail.

jdddog: It might be more appropriate to use a TypeError, e.g. https://stackoverflow.com/questions/59349532/what-is-the-standard-exception-for-a-missing-value-in-python

jdddog (changes requested):
It looks good to me Aniek, aside from changing the Exception type, it just needs to be rebased and the conflict in test_utils.py to be resolved.

jdddog (approved):

PR #408, opened 2021-04-29
Update telescope template BigQuery load methods

aroelo:
* Add function to load partition (based on table field, not ingestion date)
* Update mellon telescopes to load data into partition instead of shard
* Add unit tests


aroelo:
@rhosking @jdddog 
I've just started with Google Analytics as an example to load data into a partition. Some things to think about/discuss:

- Do we want to add a specific field, like 'partition_date' to the data, so this is the same for all partitioned telescopes? 
The alternative is just to use the relevant field directly, e.g. for google analytics this could be the 'start_date' or 'end_date' field. However, for some tables, e.g. google_books_traffic, there is no date included in the data itself, so we would have to add a partition date anyway (for the shards the date is stored in the table name).

- Do we want a default time partitioning type (e.g. month)?

- Should we change the default 'require_partition_filter' in load_bigquery_table to 'False', instead of 'True'? It seems that for most uses so far this value is set to False, so perhaps it makes sense to set this as the default value.

- I would like to add the 'bq_load_partition' method that is now implemented in the Google Analytics telescope to the SnapshotTelescope, but this will have to wait until we have updated the template to be able to get the project id & dataset location (coming from the organisation)

Please add any other points you would like to discuss.

rhosking:
> * Do we want to add a specific field, like 'partition_date' to the data, so this is the same for all partitioned telescopes?
>   The alternative is just to use the relevant field directly, e.g. for google analytics this could be the 'start_date' or 'end_date' field. However, for some tables, e.g. google_books_traffic, there is no date included in the data itself, so we would have to add a partition date anyway (for the shards the date is stored in the table name).

For the Mellon Telescopes, I believe most are specifically fetched for a particular month, so I think it is reasonable for a field to be added that represents this. Just to expand on this, I believe most of the data is at the granularity of the month explicitly because of the way we request it. We could request it in different units of aggregation, but we don't. So I'm all for having this explicit month field in those datasets, it will actually help the downstream analysis too. But when looking at telescopes like crossref events, using an existing field makes sense, as this data does not arrive aggregated.

> Do we want a default time partitioning type (e.g. month)?

I'm not sure about the default for that, I think it will be different for every telescope, so will almost always be explicitly set by whatever is calling the method. 

> Should we change the default 'require_partition_filter' in load_bigquery_table to 'False', instead of 'True'? It seems that for most uses so far this value is set to False, so perhaps it makes sense to set this as the default value.

Yes, I think that makes sense. That will be the more common case of usage. It's more flexible with it set to False, you get all of the benefits when you do use the partition as part of a query, but you are not forced to do so. Forcing a True is only really relevant when dealing with very very large amounts of data, and you don't want to allow end users to accidentally spend too much month.

Just a FYI, I added support for clustering to the load_bigquery_table in this pull request: https://github.com/The-Academic-Observatory/observatory-platform/commit/724e466fb2a7ae9baa8aeb05405fabec770f92a4

jdddog (commented):
> * Do we want to add a specific field, like 'partition_date' to the data, so this is the same for all partitioned telescopes?
>   The alternative is just to use the relevant field directly, e.g. for google analytics this could be the 'start_date' or 'end_date' field. However, for some tables, e.g. google_books_traffic, there is no date included in the data itself, so we would have to add a partition date anyway (for the shards the date is stored in the table name).

If it is feasible to add a specific field to each telescope then that would be good. Do you think that this would be complicated?

> * Do we want a default time partitioning type (e.g. month)?

I think it would be better to do day partitioning, because you can always filter down to a specific DAG run and you will know the specific date that is being referred to (which you won't if it is month based). Also, if we changed our minds and decided to change DAG frequency to fortnightly, weekly or daily then a monthly partition would lump these all together.

> * Should we change the default 'require_partition_filter' in load_bigquery_table to 'False', instead of 'True'? It seems that for most uses so far this value is set to False, so perhaps it makes sense to set this as the default value.

Yes that makes sense. Just check that the existing uses that have it set to True implicitly are changed to explicitly set it to True.

What is the purpose of having the main table and the partition table?

aroelo:
> > * Do we want to add a specific field, like 'partition_date' to the data, so this is the same for all partitioned telescopes?
> >   The alternative is just to use the relevant field directly, e.g. for google analytics this could be the 'start_date' or 'end_date' field. However, for some tables, e.g. google_books_traffic, there is no date included in the data itself, so we would have to add a partition date anyway (for the shards the date is stored in the table name).
> 
> If it is feasible to add a specific field to each telescope then that would be good. Do you think that this would be feasible or complicated?

Do you mean for all the telescopes or at first for the Mellon telescopes? For the Mellon telescopes this is feasible. 
Most of them have a 'results' list with dictionaries, we can then simply add an additional key/value pair to each dictionary in the list, using [this](https://github.com/The-Academic-Observatory/observatory-platform/blob/telescopes_write_partition/observatory-platform/observatory/platform/utils/template_utils.py#L131-L148). 
The only exception is oapen irus uk, there we download a blob that has the already transformed data. We could either load the data from the file into a list and similarly add the key/value pairs and then write it back to a file. Or we could add the key/value pairs directly during the Cloud Function [here](https://github.com/The-Academic-Observatory/oapen-irus-uk-cloud-function/blob/3a1da1c2f560a0a970881022304991b76e3d3b56/main.py#L258))

If you mean for all telescopes in general I need to have a closer look.

> > * Do we want a default time partitioning type (e.g. month)?
> 
> I think it would be better to do day partitioning, because you can always filter down to a specific DAG run and you will know the specific date that is being referred to (which you won't if it is month based). Also, if we changed our minds and decided to set the DAG frequency to fortnightly, weekly or daily then a monthly partition would lump these all together.

Yeah that makes sense. I can set it to daily for the Mellon telescopes. Do you think a default in general would still be good or rather just set it to daily explicitly?

> What is the purpose of having the main table and the partition table?

Do you mean in the StreamTelescopes? The main/partition table set up is not used for any of the Mellon telescopes, since they are all 'Snapshots'. 
For the StreamTelescopes in first instance all data is loaded into the main table. Then when we get newer releases, there might be some entries with the same id that got edited/deleted.
Each following release is initially loaded in a separate partition in the '_partitions' table. Then, the partitions are merged with the main table and the old entries are replaced with the new entries.

For the Mellon telescopes it is different as there won't be any entries with the same id that will be edited or deleted, each entry is unique.

rhosking:
> Yeah that makes sense. I can set it to daily for the Mellon telescopes. Do you think a default in general would still be good or rather just set it to daily explicitly?

What happens when you set it too daily, but the input dates are monthly? Most of Mellon data we are getting it is at the monthly frequency (aka one row = one month). I think mostly I need to better understand how this works: https://cloud.google.com/bigquery/docs/partitioned-tables#daily_partitioning_versus_hourly_monthly_or_yearly_partitioning

Daily def makes sense for crossref events.

aroelo:
> > Yeah that makes sense. I can set it to daily for the Mellon telescopes. Do you think a default in general would still be good or rather just set it to daily explicitly?
> 
> What happens when you set it too daily, but the input dates are monthly? Most of Mellon data we are getting it is at the monthly frequency (aka one row = one month). I think mostly I need to better understand how this works: https://cloud.google.com/bigquery/docs/partitioned-tables#daily_partitioning_versus_hourly_monthly_or_yearly_partitioning
> 
> Daily def makes sense for crossref events.

There's two points where the input date matters and has to match the partition type.

-  The first is the field you use for partitioning. This has to match the partition type, it can just be a DATE type for most partitioning types, and you need a more granular TIMESTAMP or DATETIME if you want e.g. hourly partitioning.

For the Mellon telescopes we will use the release date, this is a datetime object so you can just get the 'YYYY-MM-DD' format (e.g. https://github.com/The-Academic-Observatory/observatory-platform/blob/e775431fe7befff6bee8bc19d58f43884f560859/observatory-platform/observatory/platform/utils/template_utils.py#L131-L148). 
For most of them there is not a single release date, because we're collecting data over a period, so I usually just use the end date of the period as the 'release date'. It would probably be good to standardise this.

From Google:
> If the table is partitioned on a DATE column, you can create partitions with either daily, monthly, or yearly granularity. Each partition contains a range of values where the start of the range is the beginning of a day, month, or year, and the interval of the range is one day, month, or year based on the partitioning granularity. If the table is partitioned on a TIMESTAMP or DATETIME column, you can create partitions with any time-unit granularity type, including HOUR.

- The second is in the naming of the table. I found that if you use the e.g. 'table$YYYYMMDD' format, you can actually load the data into a single partition and won't have that issue of needing to set it to append to the table. (Note this is slightly different to the naming format of shards 'tableYYYYMMDD'
This format needs to match the partitioning type. If you set the type to e.g. monthly, the table name needs to be 'table$YYYYMM', if it is set to daily it should be 'table%YYYYMMDD'.

You can find that in the 'partition identifiers' bit in the table here: 
https://cloud.google.com/bigquery/docs/partitioned-tables#dt_compare

I've added it here to make the correct table id:
https://github.com/The-Academic-Observatory/observatory-platform/blob/e775431fe7befff6bee8bc19d58f43884f560859/observatory-platform/observatory/platform/utils/template_utils.py#L168-L188

observatory-platform/observatory/platform/utils/template_utils.py
```diff
@@ -46,6 +40,8 @@
                                                  upload_files_to_cloud_storage)
 from observatory.platform.utils.jinja2_utils import make_sql_jinja2_filename, render_template
 
+partition_field = 'partition_date'
```
jdddog, line 43: This should be a parameter in each function rather than a constant, as you may want to change what field is the partition for a specific table.

observatory-platform/observatory/platform/utils/template_utils.py
```diff
@@ -122,13 +122,37 @@ def upload_files_from_list(files_list: List[str], bucket_name: str) -> bool:
...
+    :param partition_type: The partition type
+    :return: Updated list of dicts with partition dates
+    """
+    if partition_type == bigquery.TimePartitioningType.HOUR:
+        partition_date = partition_date.isoformat()
+    else:
+        partition_date = partition_date.strftime('%Y-%m-%d')
+
+    for entry in list_of_dicts:
+        entry[partition_field] = partition_date
```
jdddog, line 147: This should come from a parameter in the function, as you may want to change what field is the partition for a specific table.

observatory-platform/observatory/platform/utils/template_utils.py
```diff
@@ -274,29 +322,73 @@ def bq_load_shard_v2(project_id: str, transform_bucket: str, transform_blob: str
...
+    schema_file_path = prepare_bq_load_v2(project_id, dataset_id, dataset_location, table_id, release_date,
+                                          prefix,
+                                          schema_version, dataset_description)
+
+    uri = f"gs://{transform_bucket}/{transform_blob}"
+
+    # Include date in table id, so data in table is not overwritten
+    table_id = create_date_table_id(table_id, release_date, partition_type)
+    success = load_bigquery_table(uri, dataset_id, dataset_location, table_id, schema_file_path,
+                                  source_format, partition=True, partition_field=partition_field,
```
jdddog, line 390: This should come from a parameter in the function, as you may want to change what field is the partition for a specific table.

jdddog (changes requested):
It looks good Aniek, just a few comments around standardising the tables:

> "For most of them there is not a single release date, because we're collecting data over a period, so I usually just use the end date of the period as the 'release date'. It would probably be good to standardise this."

I think it would be good to use the end of the period, however it looks like the release date is being set to to the start date of the period. For instance, with the Google Books telescope, the google_books_sales `partition_date` field is set to 2021-02-01 when processing February 2021, when it should be set to 2021-02-28. So the day is set to the start of the month.

It also looks like Google Books has changed the date format, to month/day/year i.e. 2/13/21 `ERROR - time data '2/1/21' does not match format '%d/%m/%Y'` so the format needs to be set to `'%m/%d/%y'`.

Also, sometimes the name of the file being processed has a different date (which sets the release date) to the rows contained within the file. So it might be good to have a sanity check to make sure that the dates of the processed rows fall within the month being processed.

Instead of calling the partition field `partition_date`, it would be more descriptive to call it `release_date` (and to adjust the description accordingly, stating that it is the last day of the month being processed.

It would also be good to sort the data by date when applicable, before it is loaded into BigQuery (from earliest to most recent). For example, the data loaded into the Google Books `google_books_sales` could be sorted by the `Transaction_Date` column before being loaded.

aroelo:
> It looks good Aniek, just a few comments around standardising the tables:

Thanks for the review :-).

> > "For most of them there is not a single release date, because we're collecting data over a period, so I usually just use the end date of the period as the 'release date'. It would probably be good to standardise this."
> 
> I think it would be good to use the end of the period, however it looks like the release date is being set to to the start date of the period. For instance, with the Google Books telescope, the google_books_sales `partition_date` field is set to 2021-02-01 when processing February 2021, when it should be set to 2021-02-28. So the day is set to the start of the month.

Ah I see, I think for most telescopes when there is a start date and end date I set the 'release date' to the end date. For google books there is no start and end date, but just 1 date from the file name in the `'%Y_%m'` format. When it is parsed by pendulum it defaults to the start of that month. I will change it to the end of the month, as well as for the other telescopes that work similarly.

> It also looks like Google Books has changed the date format, to month/day/year i.e. 2/13/21 `ERROR - time data '2/1/21' does not match format '%d/%m/%Y'` so the format needs to be set to `'%m/%d/%y'`.

Mmm yes, I thought I fixed it in the last commit, but apparently still set it wrong. I've changed it to `'%m/%d/%y'`. Thanks for picking that up.

> Also, sometimes the name of the file being processed has a different date (which sets the release date) to the rows contained within the file. So it might be good to have a sanity check to make sure that the dates of the processed rows fall within the month being processed.

Sure, I could add a check for the sales report, but for the traffic report this is not possible. So the date in the file name really needs to be correct. There is a risk of manual error here.

> Instead of calling the partition field `partition_date`, it would be more descriptive to call it `release_date` (and to adjust the description accordingly, stating that it is the last day of the month being processed.

I hardcoded the name 'partition_date' as I thought it might be useful to have this the same in each table. The partition date can then point to something different for each telescope. The release date makes the most sense for these, but perhaps for other telescopes the partition date can be something different than the release date.

For example if for the google books telescope we would like to partition on the transaction date instead of the release date you could still add the field 'partition_date' and copy those values of the transaction date, instead of directly partitioning on the 'transaction date' field. It might make it easier downstream to query tables if you know that the table is always partitioned on 'partition_date'.

I will make the changes to add the field name as a parameter though in the functions where it is used and set the default to 'release_date'.

> It would also be good to sort the data by date when applicable, before it is loaded into BigQuery (from earliest to most recent). For example, the data loaded into the Google Books `google_books_sales` could be sorted by the `Transaction_Date` column before being loaded.

Could this be solved by clustering in BigQuery instead? I know @rhosking is interested in using clustering to the current telescopes. It might be a bit redundant to sort the data before it goes into BigQuery and then sort it again once it is in BigQuery.

I've also updated the JSTOR telescope as the end date was not set correctly for older reports that were created manually.

observatory-dags/observatory/dags/telescopes/google_books.py
```diff
@@ -107,9 +111,19 @@ def transform(self):
                 for row in csv_reader:
                     transformed_row = OrderedDict((convert(k.replace('%', 'Perc')), v) for k, v in row.items())
                     if 'sales' in file:
+                        transaction_date = datetime.strptime(transformed_row['Transaction_Date'], '%m/%d/%Y')
```
jdddog, line 114: This format string should be `%-d/%-m/%-y` to match the new date format that Google Books is using, e.g. 1/1/18

observatory-dags/observatory/dags/telescopes/oapen_irus_uk.py
```diff
@@ -40,8 +42,9 @@
...
     download_blob_from_cloud_storage, \
     upload_file_to_cloud_storage
-from observatory.platform.utils.telescope_utils import make_dag_id, make_org_id
-from observatory.platform.utils.template_utils import SubFolder, blob_name, telescope_path
+from observatory.platform.utils.telescope_utils import make_dag_id, make_org_id, add_partition_date, list_to_jsonl_gz
+from observatory.platform.utils.template_utils import SubFolder, blob_name, telescope_path, bq_load_partition, \
+    table_ids_from_path
 
 
 class OapenIrusUkRelease(SnapshotRelease):
```
jdddog, line 50: The cloud function gets the following error ` download_access_stats_new doi = item['Item_ID']['DOI'] KeyError: 'DOI'`:
```
_function_handler.invoke_user_function(flask.request) File "/env/local/lib/python3.7/site-packages/google/cloud/functions/worker_v2.py", line 268, in invoke_user_function return call_user_function(request_or_event) File "/env/local/lib/python3.7/site-packages/google/cloud/functions/worker_v2.py", line 261, in call_user_function return self._user_function(request_or_event) File "/user_code/main.py", line 62, in download download_access_stats_new(file_path, release_date, username, password, publisher_uuid, geoip_client) File "/user_code/main.py", line 219, in download_access_stats_new doi = item['Item_ID']['DOI'] KeyError: 'DOI'
Traceback (most recent call last): File "/env/local/lib/python3.7/site-packages/google/cloud/functions/worker_v2.py", line 402, in run_http_function result = _function_handler.invoke_user_function(flask.request) File "/env/local/lib/python3.7/site-packages/google/cloud/functions/worker_v2.py", line 268, in invoke_user_function return call_user_function(request_or_event) File "/env/local/lib/python3.7/site-packages/google/cloud/functions/worker_v2.py", line 261, in call_user_function return self._user_function(request_or_event) File "/user_code/main.py", line 62, in download download_access_stats_new(file_path, release_date, username, password, publisher_uuid, geoip_client) File "/user_code/main.py", line 219, in download_access_stats_new doi = item['Item_ID']['DOI'] KeyError: 'DOI'
```

jdddog (commented):
Hey Aniek, the partition changes look good. There are a couple of problems with Google Books and OApen which need fixing though (see the comments).

observatory-dags/observatory/dags/telescopes/google_books.py
aroelo: Ah I see. I updated the test file to match the non-zero date formats. However, the month is before the day, so then it would be `%-m/%-d%/%-y`.
This does not work on mac OS though, I tested with `%m/%d/%y` and that does seem to pick up the 1 decimal months/days correctly.

observatory-dags/observatory/dags/telescopes/oapen_irus_uk.py
aroelo: Mmm, apparently there are some items without a DOI or some of the other identifiers, I was not aware of this.
I've added a try/except block to get these identifiers in the cloud function. Will have to create a new release though and update this in the telescope.

observatory-dags/observatory/dags/telescopes/google_books.py
jdddog: Ah right, that makes sense

jdddog (changes requested):
There are a couple more errors.

For Google Books it fails to load into BigQuery, because it can't find a schema, as the schema date is too recent:
```python
[2021-05-24 22:02:43,966] {template_utils.py:138} INFO - Creating table ids from path: /opt/observatory/data/telescopes/transform/google_books_anu_press/google_books_anu_press_2018_01_31/google_books_sales.jsonl.gz
[2021-05-24 22:02:43,966] {template_utils.py:143} INFO - Table id: google_books_sales, partition table id: google_books_sales_partitions
[2021-05-24 22:02:43,976] {gc_utils.py:133} INFO - create_bigquery_dataset: creating dataset dataset_ref=coki-jamie-dev.google
[2021-05-24 22:02:45,174] {config_utils.py:116} INFO - Looking for schema with search parameters: analysis_schema_path=/opt/observatory-dags/observatory/dags/database/schema, prefix=, table_name=google_books_sales, release_date=2018-01-31T23:59:59.999999+00:00, version=None
[2021-05-24 22:02:45,175] {config_utils.py:167} ERROR - No schema found.
[2021-05-24 22:02:48,768] {local_task_job.py:102} INFO - Task exited with return code 1
```
I think ideally the schema would not have a date, perhaps just a version. I don't think the schema finding function supports this at the moment, so perhaps set a really early date (e.g. 1900-01-1. We can modify the schema finding function later.

The Oapen telescope fails to load data into BigQuery as well, after the changes to the Cloud Function:
```python
[2021-05-24 22:02:55,267] {gc_utils.py:203} INFO - load_bigquery_table: load bigquery table uri=gs://coki-jamie-dev-transform-7acdedff/telescopes/oapen_irus_uk_anu_press/oapen_irus_uk_anu_press_2020_04_30/oapen_irus_uk.jsonl.gz, dataset_id=oapen, location=us, table=oapen_irus_uk$202004, schema_file_path=/opt/observatory-dags/observatory/dags/database/schema/oapen_irus_uk_2020-04-01.json, source_format=NEWLINE_DELIMITED_JSON
[2021-05-24 22:02:57,080] {gc_utils.py:250} ERROR - load_bigquery_table: load bigquery table failed: 400 Provided Schema does not match Table coki-jamie-dev:oapen.oapen_irus_uk$202004. Cannot add fields (field: URI).
[2021-05-24 22:02:57,081] {gc_utils.py:252} ERROR - Error collection:
[{'reason': 'invalid', 'message': 'Provided Schema does not match Table coki-jamie-dev:oapen.oapen_irus_uk$202004. Cannot add fields (field: URI)'}]
[2021-05-24 22:02:57,087] {taskinstance.py:1150} ERROR - 
  File "/home/airflow/.local/lib/python3.7/site-packages/airflow/models/taskinstance.py", line 984, in _run_raw_task
    result = task_copy.execute(context=context)
  File "/home/airflow/.local/lib/python3.7/site-packages/airflow/operators/python_operator.py", line 113, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.7/site-packages/airflow/operators/python_operator.py", line 118, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/opt/observatory-platform/observatory/platform/telescopes/telescope.py", line 306, in task_callable
    result = func(release, **kwargs)
  File "/opt/observatory-dags/observatory/dags/telescopes/oapen_irus_uk.py", line 330, in bq_load_partition
    table_description=table_description, **self.load_bigquery_table_kwargs)
  File "/opt/observatory-platform/observatory/platform/utils/template_utils.py", line 374, in bq_load_partition
    raise AirflowException()
airflow.exceptions.AirflowException
```

aroelo:
> There are a couple more errors.
> 
> For Google Books it fails to load into BigQuery, because it can't find a schema, as the schema date is too recent:
> 
> ```python
> [2021-05-24 22:02:43,966] {template_utils.py:138} INFO - Creating table ids from path: /opt/observatory/data/telescopes/transform/google_books_anu_press/google_books_anu_press_2018_01_31/google_books_sales.jsonl.gz
> [2021-05-24 22:02:43,966] {template_utils.py:143} INFO - Table id: google_books_sales, partition table id: google_books_sales_partitions
> [2021-05-24 22:02:43,976] {gc_utils.py:133} INFO - create_bigquery_dataset: creating dataset dataset_ref=coki-jamie-dev.google
> [2021-05-24 22:02:45,174] {config_utils.py:116} INFO - Looking for schema with search parameters: analysis_schema_path=/opt/observatory-dags/observatory/dags/database/schema, prefix=, table_name=google_books_sales, release_date=2018-01-31T23:59:59.999999+00:00, version=None
> [2021-05-24 22:02:45,175] {config_utils.py:167} ERROR - No schema found.
> [2021-05-24 22:02:48,768] {local_task_job.py:102} INFO - Task exited with return code 1
> ```
> 
> I think ideally the schema would not have a date, perhaps just a version. I don't think the schema finding function supports this at the moment, so perhaps set a really early date (e.g. 1900-01-01. We can modify the schema finding function later.

Ah yep, that's a good idea. I'll change the date in the schema name.

> The Oapen telescope fails to load data into BigQuery as well, after the changes to the Cloud Function:
> 
> ```python
> [2021-05-24 22:02:55,267] {gc_utils.py:203} INFO - load_bigquery_table: load bigquery table uri=gs://coki-jamie-dev-transform-7acdedff/telescopes/oapen_irus_uk_anu_press/oapen_irus_uk_anu_press_2020_04_30/oapen_irus_uk.jsonl.gz, dataset_id=oapen, location=us, table=oapen_irus_uk$202004, schema_file_path=/opt/observatory-dags/observatory/dags/database/schema/oapen_irus_uk_2020-04-01.json, source_format=NEWLINE_DELIMITED_JSON
> [2021-05-24 22:02:57,080] {gc_utils.py:250} ERROR - load_bigquery_table: load bigquery table failed: 400 Provided Schema does not match Table coki-jamie-dev:oapen.oapen_irus_uk$202004. Cannot add fields (field: URI).
> [2021-05-24 22:02:57,081] {gc_utils.py:252} ERROR - Error collection:
> [{'reason': 'invalid', 'message': 'Provided Schema does not match Table coki-jamie-dev:oapen.oapen_irus_uk$202004. Cannot add fields (field: URI)'}]
> [2021-05-24 22:02:57,087] {taskinstance.py:1150} ERROR - 
> Traceback (most recent call last):
>   File "/home/airflow/.local/lib/python3.7/site-packages/airflow/models/taskinstance.py", line 984, in _run_raw_task
>     result = task_copy.execute(context=context)
>   File "/home/airflow/.local/lib/python3.7/site-packages/airflow/operators/python_operator.py", line 113, in execute
>     return_value = self.execute_callable()
>   File "/home/airflow/.local/lib/python3.7/site-packages/airflow/operators/python_operator.py", line 118, in execute_callable
>     return self.python_callable(*self.op_args, **self.op_kwargs)
>   File "/opt/observatory-platform/observatory/platform/telescopes/telescope.py", line 306, in task_callable
>     result = func(release, **kwargs)
>   File "/opt/observatory-dags/observatory/dags/telescopes/oapen_irus_uk.py", line 330, in bq_load_partition
>     table_description=table_description, **self.load_bigquery_table_kwargs)
>   File "/opt/observatory-platform/observatory/platform/utils/template_utils.py", line 374, in bq_load_partition
>     raise AirflowException()
> airflow.exceptions.AirflowException
> ```

Mmm, I tested the telescope end to end with real data after updating the function (should have done it before the key error appeared.. :-)). Could this be an error because you already have an existing table with the older schema? Because it says 'Provided Schema does not match Table coki-jamie-dev:oapen.oapen_irus_uk$202004.'

jdddog:
> > There are a couple more errors.
> > For Google Books it fails to load into BigQuery, because it can't find a schema, as the schema date is too recent:
> > ```python
> > [2021-05-24 22:02:43,966] {template_utils.py:138} INFO - Creating table ids from path: /opt/observatory/data/telescopes/transform/google_books_anu_press/google_books_anu_press_2018_01_31/google_books_sales.jsonl.gz
> > [2021-05-24 22:02:43,966] {template_utils.py:143} INFO - Table id: google_books_sales, partition table id: google_books_sales_partitions
> > [2021-05-24 22:02:43,976] {gc_utils.py:133} INFO - create_bigquery_dataset: creating dataset dataset_ref=coki-jamie-dev.google
> > [2021-05-24 22:02:45,174] {config_utils.py:116} INFO - Looking for schema with search parameters: analysis_schema_path=/opt/observatory-dags/observatory/dags/database/schema, prefix=, table_name=google_books_sales, release_date=2018-01-31T23:59:59.999999+00:00, version=None
> > [2021-05-24 22:02:45,175] {config_utils.py:167} ERROR - No schema found.
> > [2021-05-24 22:02:48,768] {local_task_job.py:102} INFO - Task exited with return code 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think ideally the schema would not have a date, perhaps just a version. I don't think the schema finding function supports this at the moment, so perhaps set a really early date (e.g. 1900-01-01. We can modify the schema finding function later.
> 
> Ah yep, that's a good idea. I'll change the date in the schema name.
> 
> > The Oapen telescope fails to load data into BigQuery as well, after the changes to the Cloud Function:
> > ```python
> > [2021-05-24 22:02:55,267] {gc_utils.py:203} INFO - load_bigquery_table: load bigquery table uri=gs://coki-jamie-dev-transform-7acdedff/telescopes/oapen_irus_uk_anu_press/oapen_irus_uk_anu_press_2020_04_30/oapen_irus_uk.jsonl.gz, dataset_id=oapen, location=us, table=oapen_irus_uk$202004, schema_file_path=/opt/observatory-dags/observatory/dags/database/schema/oapen_irus_uk_2020-04-01.json, source_format=NEWLINE_DELIMITED_JSON
> > [2021-05-24 22:02:57,080] {gc_utils.py:250} ERROR - load_bigquery_table: load bigquery table failed: 400 Provided Schema does not match Table coki-jamie-dev:oapen.oapen_irus_uk$202004. Cannot add fields (field: URI).
> > [2021-05-24 22:02:57,081] {gc_utils.py:252} ERROR - Error collection:
> > [{'reason': 'invalid', 'message': 'Provided Schema does not match Table coki-jamie-dev:oapen.oapen_irus_uk$202004. Cannot add fields (field: URI)'}]
> > [2021-05-24 22:02:57,087] {taskinstance.py:1150} ERROR - 
> > Traceback (most recent call last):
> >   File "/home/airflow/.local/lib/python3.7/site-packages/airflow/models/taskinstance.py", line 984, in _run_raw_task
> >     result = task_copy.execute(context=context)
> >   File "/home/airflow/.local/lib/python3.7/site-packages/airflow/operators/python_operator.py", line 113, in execute
> >     return_value = self.execute_callable()
> >   File "/home/airflow/.local/lib/python3.7/site-packages/airflow/operators/python_operator.py", line 118, in execute_callable
> >     return self.python_callable(*self.op_args, **self.op_kwargs)
> >   File "/opt/observatory-platform/observatory/platform/telescopes/telescope.py", line 306, in task_callable
> >     result = func(release, **kwargs)
> >   File "/opt/observatory-dags/observatory/dags/telescopes/oapen_irus_uk.py", line 330, in bq_load_partition
> >     table_description=table_description, **self.load_bigquery_table_kwargs)
> >   File "/opt/observatory-platform/observatory/platform/utils/template_utils.py", line 374, in bq_load_partition
> >     raise AirflowException()
> > airflow.exceptions.AirflowException
> > ```
> 
> Mmm, I tested the telescope end to end with real data after updating the function (should have done it before the key error appeared.. :-)). Could this be an error because you already have an existing table with the older schema? Because it says 'Provided Schema does not match Table coki-jamie-dev:oapen.oapen_irus_uk$202004.'

Ah perhaps, let me give it a try again from the beginning

jdddog:
Once I start again it works for me

jdddog (approved):

PR #409, opened 2021-05-05
ORCID telescope

### LINKED ISSUE #233: Telescope workflow implementation: ORCID
**Name**: ORCID
**Harvest Type**: Snapshot/API
**Snapshot Frequency**: Yearly (?)

A public data file is published yearly which contains a snapshot of all public data in the ORCID Registry associated with an ORCID record that was created or claimed by an individual up until that date, see https://support.orcid.org/hc/en-us/articles/360006897394-How-do-I-get-the-public-data-file-#01 and https://orcid.figshare.com/. 

> The file contains the public information associated with each user's ORCID record. Each record is included as a separate file in both JSON and XML in the 2013-2017 files. In the 2018 file, each record is included as a separate file in XML, and each full activity section for each record is included as a separate file in XML. For those who prefer JSON, use the [ORCID Conversion Library](https://github.com/ORCID/orcid-conversion-lib) to convert the XML files to JSON.

As an ORCID member (which I don't think we are?) you receive credentials to access their Amazon S3 ORCID data sync repository any time and create a snapshot.

They also have the ORCID public API which we could maybe use to get all records see e.g. https://members.orcid.org/api/tutorial/search-orcid-registry.

@rhosking would a yearly update be sufficient or would it be worthwhile to explore options to get updates more frequently?

aroelo:
The ORCID telescope transfers the summary file for each record from the AWS S3 bucket to a Google Cloud storage bucket.

- A separate GCS bucket needs to be created, because it is not possible to specify a directory inside an existing bucket for the transfer job.

- The first run will transfer all data and then download all files from the GCS bucket to the local drive so they can be transformed. I found that using gsutil is the fastest way to download all files (faster than using the python library), so I have added installation of this to the Docker file. 

- The second run (and any after that) will start a transfer job, but only transfer files that have been modified in the AWS bucket since the end date of the previous run. There is no way to get a list of which files were transferred with the transfer job, so in order to know which files need to be processed further it will use the lambda file from ORCID that is hosted in another AWS bucket.

- The lambda file contains info on when records are last modified, the records are sorted by last modified date. It will get the file names of the files modified between the end date of the previous run and the start date of the current run. These file names are stored in a seperate txt file that is used as input for the gsutil command to download the files.

- There might be some mismatch between the info in this lambda file and the actual modification times of the files in the AWS bucket. I noticed that there is e.g. a record in the lambda file with a modification datetime, but this file can't be found in the GCS bucket or in the AWS bucket. Perhaps the record has been deleted after modification? Or deleting a record counts as a 'modification'?
When a file can not be found in the GCS bucket it will raise an CommandException, so I have added a check that the missing file exceptions are allowed and will not raise an AirflowException.

- I have found that running the `download_transferred` task using an Airflow instance with a postgres server created locally with docker will give an error after ~2 hours. I'm not sure what is causing this error, there is nothing in the postgres log, the other tasks all work fine (also long tasks longer than 2 hours) and when running the bash command that is executed using subprocess directly it will finish without any problems. Perhaps this will not happen with a GCS postgres server? 
See details for the Airflow log.
<details>

```
[2021-07-05 09:01:35,681] {orcid.py:151} INFO - Executing bash command: gsutil -m -q cp -I /opt/observatory/data/telescopes/download/orcid/2021_06_24-2021_07_04
[2021-07-05 11:10:06,203] {base_job.py:202} ERROR - LocalTaskJob heartbeat got an exception
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 2336, in _wrap_pool_connect
    return fn()
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 364, in connect
    return _ConnectionFairy._checkout(self)
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 778, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 495, in checkout
    rec = pool._do_get()
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/pool/impl.py", line 241, in _do_get
    return self._create_connection()
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 309, in _create_connection
    return _ConnectionRecord(self)
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 440, in __init__
    self.__connect(first_connect_check=True)
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 661, in __connect
    pool.logger.debug("Error on connect(): %s", e)
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/util/langhelpers.py", line 70, in __exit__
    with_traceback=exc_tb,
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 656, in __connect
    connection = pool._invoke_creator(self)
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/engine/strategies.py", line 114, in connect
    return dialect.connect(*cargs, **cparams)
  File "/home/airflow/.local/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 493, in connect
    return self.dbapi.connect(*cargs, **cparams)
  File "/home/airflow/.local/lib/python3.7/site-packages/psycopg2/__init__.py", line 127, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: could not translate host name "postgres" to address: Temporary failure in name resolution
```
</details>

EDIT: The download_transferred task does run successfully using a terraform deployment (with a postgres server hosted on Google Cloud), I'm not sure what is causing the local error, but happy to consider this as fixed regarding this telescope. 

- The transformation includes reading the xml data into a dict and changing the keys in the dict so they are conforming to BQ standards. This is done for each individual file/record in parallel.
Since the BQ loading file size is limited to 4GB for gzipped files they are concatenated and compressed per group, the group is defined by the last 3 digits of the file name (this is also how the files are structured in the AWS bucket). 
These compressed files are uploaded to the transform bucket and all files in the transform bucket are then loaded into BQ at once as a batch load.

closes #233
closes #271


tests/observatory/dags/telescopes/test_orcid.py
```diff
@@ -0,0 +1,487 @@
...
+
+        # Setup Telescope
+        telescope = OrcidTelescope(dataset_id=dataset_id)
+        dag = telescope.make_dag()
+
+        # Create the Observatory environment and run tests
+        with env.create():
+            # Add connection
+            conn = Connection(conn_id=AirflowConns.ORCID,
+                              uri="aws://UWLA41aAhdja:AJLD91saAJSKAL0AjAhkaka@")
```
jdddog, line 114: I presume that these are not real credentials?

jdddog (commented):
Hey Aniek, it looks good to me, just a couple of things:
* The credentials in the tests aren't real? (see comment).
* Could you add a [sensor to the DOI workflow](https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory-dags/observatory/dags/workflows/doi_workflow.py#L268) so that it waits for the ORCID telescope to finish when it runs every week? You'll also need to update the DOI Workflow test `test_dag_structure`.

And then rebase afterwards.

tests/observatory/dags/telescopes/test_orcid.py
aroelo: No they are not :-) I generated them with canarytoken, just trying that out, so you get a notification when someone tries to use the token. It does not really serve any purpose for us though, because they are publicly available obviously. The idea is you can use the tokens to check whether there is a security fault somewhere, if you expect the token to be kept private.

aroelo:
> Hey Aniek, it looks good to me, just a couple of things:
> 
> * The credentials in the tests aren't real? (see comment).
> * Could you add a [sensor to the DOI workflow](https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory-dags/observatory/dags/workflows/doi_workflow.py#L268) so that it waits for the ORCID telescope to finish when it runs every week? You'll also need to update the DOI Workflow test `test_dag_structure`.
> 
> And then rebase afterwards.

Thanks for the review Jamie, I've added the sensor and rebased.

jdddog (approved):

jdddog:
> > Hey Aniek, it looks good to me, just a couple of things:
> > 
> > * The credentials in the tests aren't real? (see comment).
> > * Could you add a [sensor to the DOI workflow](https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory-dags/observatory/dags/workflows/doi_workflow.py#L268) so that it waits for the ORCID telescope to finish when it runs every week? You'll also need to update the DOI Workflow test `test_dag_structure`.
> > 
> > And then rebase afterwards.
> 
> Thanks for the review Jamie, I've added the sensor and rebased.

Thanks Aniek!

tests/observatory/dags/telescopes/test_orcid.py
jdddog: Ah I see, that sounds helpful!

PR #412, opened 2021-05-11
Update readthedocs files

aroelo:
Some minor fixes to the readthedocs files of the telescopes.
Also added the 'sphinx_markdown_tables' extension so the tables are rendered correctly.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,8 +34,7 @@ extensions = [
     'sphinx.ext.intersphinx',
     'autoapi.extension',
     'recommonmark',
-    'sphinxcontrib.openapi',
-    'sphinx_markdown_tables'
+    'sphinxcontrib.openapi'
 ]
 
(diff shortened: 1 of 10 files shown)


jdddog (changes requested):
Unfortunately  `sphinx_markdown_tables` has a GPLv3 license so we can't use it.

For the tables, for now, instead you can use `eval_rst` and use rst tables, see below for an example:

\```eval_rst
insert your rst table here
\```

aroelo:
> Unfortunately `sphinx_markdown_tables` has a GPLv3 license so we can't use it.
> 
> For the tables, for now, instead you can use `eval_rst` and use rst tables, see below for an example:
> 
> ```eval_rst
> insert your rst table here
> ```

Ah alright thanks. I did not realise that.
I'll try out the rst tables :-)

jdddog (commented):
It looks good Aniek, it just needs to be rebased.

jdddog (approved):

PR #418, opened 2021-05-17
Update crossref metadata

aroelo:
Update telescope to use the template and update unit tests to run in airflow test environment.
Also changed the schedule from weekly to monthly.

### DIFF
--- a/observatory-dags/observatory/dags/telescopes/crossref_metadata.py
+++ b/observatory-dags/observatory/dags/telescopes/crossref_metadata.py
@@ -14,289 +14,228 @@
 
 # Author: Aniek Roelofs, James Diprose
 
-from __future__ import annotations
 
+from __future__ import annotations, annotations
+
+import datetime
 import functools
-import glob
 import logging
 import os
-import pathlib
 import shutil
 import subprocess
-import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
+from datetime import datetime
 from multiprocessing import cpu_count
 from subprocess import Popen
+from typing import Dict, List
 
-import pendulum
 import requests
 from airflow.exceptions import AirflowException
 from airflow.hooks.base_hook import BaseHook
-from airflow.models.taskinstance import TaskInstance
-from google.cloud.bigquery import SourceFormat
 from natsort import natsorted
-
-from observatory.dags.config import schema_path
-from observatory.platform.utils.airflow_utils import AirflowConns, \
-    AirflowVariable as Variable, \
-    AirflowVars, \
-    check_connections, \
-    check_variables
-from observatory.platform.utils.config_utils import (find_schema)
-from observatory.platform.utils.gc_utils import (bigquery_sharded_table_id,
-                                                 bigquery_table_exists,
-                                                 create_bigquery_dataset,
-                                                 load_bigquery_table,
-                                                 upload_file_to_cloud_storage,
-                                                 upload_files_to_cloud_storage)
+from observatory.platform.telescopes.snapshot_telescope import SnapshotRelease, SnapshotTelescope
+from observatory.platform.utils.airflow_utils import AirflowConns, AirflowVars
 from observatory.platform.utils.proc_utils import wait_for_process
-from observatory.platform.utils.template_utils import SubFolder, telescope_path, test_data_path
+from observatory.platform.utils.template_utils import upload_files_from_list
 from observatory.platform.utils.url_utils import retry_session
+from pendulum import Pendulum
 
 
-def download_release(release: CrossrefMetadataRelease, api_token: str):
-    """ Downloads release
+class CrossrefMetadataRelease(SnapshotRelease):
+    def __init__(self, dag_id: str, release_date: Pendulum):
+        """ Create a CrossrefMetadataRelease instance.
 
-    :param release: Instance of CrossrefRelease class
-    :param api_token: token used to access crossref data
-    :return:  None.
-    """
-
-    logging.info(f"Downloading from url: {release.url}")
-
-    # Set API token header
-    header = {
-        'Crossref-Plus-API-Token': f'Bearer {api_token}'
-    }
-
-    # Download release
-    with requests.get(release.url, headers=header, stream=True) as response:
-        # Check if authorisation with the api token was successful or not, raise error if not successful
-        if response.status_code != 200:
-            raise ConnectionError(f"Error downloading file {release.url}, status_code={response.status_code}")
-
-        # Open file for saving
-        with open(release.download_path, 'wb') as file:
-            response.raw.read = functools.partial(response.raw.read, decode_content=True)
-            shutil.copyfileobj(response.raw, file)
-
-    logging.info(f"Successfully download url to {release.download_path}")
-
-
-def extract_release(release: CrossrefMetadataRelease) -> bool:
-    """ Extract release.
-
-    :param release: Instance of CrossrefRelease class
-    """
-    logging.info(f"extract_release: {release.download_path}")
-
-    # Make directories
-    os.makedirs(release.extract_path, exist_ok=True)
-
-    # Was having issues with the next process not finding the folder that was just created, so sleep for 30 secs
-    time.sleep(30)
-
-    # Run command
-    cmd = f'tar -xv -I "pigz -d" -f {release.download_path} -C {release.extract_path}'
-    p: Popen = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, executable='/bin/bash')
-    stdout, stderr = wait_for_process(p)
-    logging.debug(stdout)
-    success = p.returncode == 0 and 'error' not in stderr.lower()
-
-    if success:
-        logging.info(f"extract_release success: {release.download_path}")
-    else:
-        logging.error(f"extract_release error: {release.download_path}")
-        logging.error(stdout)
-        logging.error(stderr)
-
-    return success
-
-
-def transform_file(input_file_path: str, output_file_path: str) -> bool:
-    r""" Transform Crossref Metadata file.
-
-    :param input_file_path: the path of the file to transform.
-    :param output_file_path: where to save the transformed file.
-    :return: whether the transformation was successful or not.
-    """
-
-    cmd = 'mawk \'BEGIN {FS="\\":";RS=",\\"";OFS=FS;ORS=RS} {for (i=1; i<=NF;i++) if(i != NF) gsub("-", "_", $i)}1\'' \
-          f' {input_file_path} | ' \
-          'mawk \'!/^\}$|^\]$|,\\"$/{gsub("\[\[", "[");gsub("]]", "]");gsub(/,[ \\t]*$/,"");' \
-          'gsub("\\"timestamp\\":_", "\\"timestamp\\":");gsub("\\"date_parts\\":\[null]", "\\"date_parts\\":[]");' \
-          'gsub(/^\{\\"items\\":\[/,"");print}\' > ' \
-          f'{output_file_path}'
-    p: Popen = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, executable='/bin/bash')
-    stdout, stderr = wait_for_process(p)
-    logging.debug(stdout)
-    success = p.returncode == 0
-
-    if success:
-        logging.info(f"transform_file success: {input_file_path}")
-    else:
-        logging.error(f"transform_file error: {input_file_path}")
-        logging.error(stderr)
-
-    return success
-
-
-def transform_release(release: CrossrefMetadataRelease, max_workers: int = cpu_count()) -> bool:
-    """ Transform a Crossref Metadata release into a form that can be loaded into BigQuery.
-
-    :param release: the CrossrefMetadataRelease release
-    :param max_workers: the number of processes to use when transforming files (one process per file).
-    :return: whether the transformation was successful or not.
-    """
-
-    # input_release_path: the path to the folder containing the extracted files for the Crossref Metadata release.
-    # output_release_path: the path where the transformed files will be saved.
-    input_release_path = release.extract_path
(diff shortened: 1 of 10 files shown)


observatory-dags/observatory/dags/telescopes/crossref_metadata.py
```diff
@@ -14,289 +14,210 @@
...
-    TASK_ID_TRANSFORM = f"transform_releases"
-    TASK_ID_UPLOAD_TRANSFORMED = 'upload_transformed'
-    TASK_ID_BQ_LOAD = f"bq_load"
-    TASK_ID_CLEANUP = f"cleanup"
 
-    @staticmethod
-    def check_dependencies(**kwargs):
-        """ Check that all variables exist that are required to run the DAG.
+    def __init__(self, dag_id: str = DAG_ID, start_date: datetime = datetime(2020, 6, 1),
+                 schedule_interval: str = '@monthly', dataset_id: str = 'crossref',
```
jdddog, line 160: The original start_date was set to `datetime(2020, 6, 7)`, so we need to keep it the same because Airflow doesn't let you change the start date once the DAG has already run.

The reason why it is set to the 7th day of the month is because of when Crossref Metadata snapshots are released: 
> "New snapshots are created each month, available by the 5th day, providing all records up to and including the previous month."

See here: https://www.crossref.org/documentation/metadata-plus/metadata-plus-snapshots/#00344

I'm not actually sure how well they keep to that schedule though, I think we should check.

If we set the `schedule_interval` to `monthly`, then that will affect the DOI workflow, which runs weekly and has a sensor that waits for the Crossref Metadata DAG to finish. I think that the sensor will wait indefinitely on weeks when the Crossref Metadata DAG does not run. 

Would you be able to find out how you can get the Crossref Metadata sensor in the DOI DAG to only wait on weeks when the Crossref Metadata DAG runs?

jdddog (changes requested):
It looks good Aniek, I've just left a few comments about scheduling and making this work in time with the DOI DAG.

observatory-dags/observatory/dags/telescopes/crossref_metadata.py
aroelo: > The original start_date was set to `datetime(2020, 6, 7)`, so we need to keep it the same because Airflow doesn't let you change the start date once the DAG has already run.
> 
> The reason why it is set to the 7th day of the month is because of when Crossref Metadata snapshots are released:
> 
> > "New snapshots are created each month, available by the 5th day, providing all records up to and including the previous month."
> 
> See here: https://www.crossref.org/documentation/metadata-plus/metadata-plus-snapshots/#00344
 
I see, I have changed this back :-)

> I'm not actually sure how well they keep to that schedule though. Could we add some logging to the check_release task that prints out all of the available releases?

I'll have a look into adding some logging to the check_release task. I think we will have to get the HEAD of each URL to check if it exists, maybe we could limit it to 2 months before and 2 months after? I don't think there is a single URL with an overview all available releases.

> I think that there is a bug in the existing code on the develop branch, because the execution date is being used to fetch the the Crossref Metadata release. Should it be execution date - 1 month? E.g. the 2021-05 release is uploaded on 2021-06-05, so if the execution date was 2021-06-01, then it would never fetch the 2021-05 release. I think that is a problem because I set the DAG to run weekly. That wouldn't be a problem for monthly though right? Because the execution date would be for the previous month?

Yeah I think you are right. This will be solved if we set it to '@monthly' and use the 'prev_execution_date' to create the URL. In that case for the execution_date 2021-06-01, it will be running on 2021-07-01 and it will check for the release of 2021-05. 
Alternatively we could set the schedule to '0 0 7 * *' (so it is after the 5th day) and use the 'execution_date' to create the URL. In that case for the execution_date of 2021-06-07, it will be running on 2021-07-07 and it will check for the release of 2021-06. So we would be a bit more up date.

> If we set the `schedule_interval` to `monthly`, then that will affect 1) the DOI workflow, which runs weekly and has a sensor that waits for the Crossref Metadata DAG to finish. I think that the sensor will wait indefinitely on weeks when the Crossref Metadata DAG does not run. 2) The worker VM is required,
> 
> Would you be able to find out how you can get the Crossref Metadata sensor in the DOI DAG to only wait on weeks when the Crossref Metadata DAG runs?

Yeah by default the ExternalTaskSensor expects the schedule_interval+start date to be in sync. We could change this by using the 'execution_date_fn' parameter of the ExternalTaskSensor. I have added a solution in the latest commit.
What do you think of this solution? I have tested it with this simple DAG:  
```python
from datetime import datetime

from airflow import DAG
from airflow.operators.dummy_operator import DummyOperator
from airflow.sensors.external_task_sensor import ExternalTaskSensor
from observatory.platform.telescopes.telescope import Telescope, Release
from observatory.platform.utils.telescope_utils import get_prev_execution_date


class FakeCrossrefTelescope(Telescope):
    DAG_ID = 'fake_crossref'
    SCHEDULE_INTERVAL = '@monthly'

    def __init__(self, dag_id: str = DAG_ID, start_date: datetime = datetime(2020, 6, 7),
                 schedule_interval: str = SCHEDULE_INTERVAL, catchup=True):
        """ Construct a Telescope instance.

        :param dag_id: the id of the DAG.
        :param start_date: the start date of the DAG.
        :param schedule_interval: the schedule interval of the DAG.
        """
        super().__init__(dag_id, start_date, schedule_interval, catchup=catchup)

        self.add_setup_task_chain([self.setup_task])
        self.add_task(self.task)

    def make_release(self, **kwargs) -> Release:
        """ Create a single release instance.

        :param kwargs: the context passed from the PythonOperator. See
        https://airflow.apache.org/docs/stable/macros-ref.html for more info.
        :return: A release instance
        """
        release_id = f"telescope_{datetime.now()}"
        release = Release(self.dag_id, release_id)
        return release

    def setup_task(self, **kwargs) -> bool:
        """ This is a custom task that is executed before a release is made. The return value is used for a
        shortcircuit operator"

        :param kwargs: the context passed from the PythonOperator. See
        https://airflow.apache.org/docs/stable/macros-ref.html for more info.
        :return: True/False
        """
        setup = 'successful'
        return True if setup == 'successful' else False

    def task(self, release: Release, **kwargs):
        """ This is a custom task that can use a release instance.

        :param release: A release instance.
        :param kwargs: the context passed from the PythonOperator. See
        https://airflow.apache.org/docs/stable/macros-ref.html for more info.
        :return: None.
        """
        print(f"This is a custom task that uses the release: {release.release_id}")


telescope = FakeCrossrefTelescope()
globals()[telescope.dag_id] = telescope.make_dag()


default_args = {
    "owner": "airflow",
    "start_date": datetime(2020, 8, 30)
}
SENSOR_ID_CROSSREF_METADATA = 'crossref_metadata_sensor'

with DAG(dag_id='test_sensor', schedule_interval='@weekly', default_args=default_args, catchup=False) as dag2:
    # Sensors
    crossref_metadata_sensor = ExternalTaskSensor(
        task_id=SENSOR_ID_CROSSREF_METADATA,
        external_dag_id=FakeCrossrefTelescope.DAG_ID,
        execution_date_fn=lambda execution_date: get_prev_execution_date(FakeCrossrefTelescope.SCHEDULE_INTERVAL,
                                                                         execution_date),
        mode='reschedule')

    dummy = DummyOperator(
        task_id='dummy'
    )

    crossref_metadata_sensor >> dummy
```

What would be the issue for the worker VM if the schedule interval is not set to '@weekly'?

jdddog: > > The original start_date was set to `datetime(2020, 6, 7)`, so we need to keep it the same because Airflow doesn't let you change the start date once the DAG has already run.
> > The reason why it is set to the 7th day of the month is because of when Crossref Metadata snapshots are released:
> > > "New snapshots are created each month, available by the 5th day, providing all records up to and including the previous month."
> > 
> > 
> > See here: https://www.crossref.org/documentation/metadata-plus/metadata-plus-snapshots/#00344
> 
> I see, I have changed this back :-)
> 

Thanks :)

> > I'm not actually sure how well they keep to that schedule though. Could we add some logging to the check_release task that prints out all of the available releases?
> 
> I'll have a look into adding some logging to the check_release task. I think we will have to get the HEAD of each URL to check if it exists, maybe we could limit it to 2 months before and 2 months after? I don't think there is a single URL with an overview all available releases.

You can find all of the years available with this URL:
`https://api.crossref.org/snapshots/monthly/`

And can query all of the releases available for a each year with this URL, e.g. for 2021: `https://api.crossref.org/snapshots/monthly/2021`

> 
> > I think that there is a bug in the existing code on the develop branch, because the execution date is being used to fetch the the Crossref Metadata release. Should it be execution date - 1 month? E.g. the 2021-05 release is uploaded on 2021-06-05, so if the execution date was 2021-06-01, then it would never fetch the 2021-05 release. I think that is a problem because I set the DAG to run weekly. That wouldn't be a problem for monthly though right? Because the execution date would be for the previous month?
> 
> Yeah I think you are right. This will be solved if we set it to '@monthly' and use the 'prev_execution_date' to create the URL. In that case for the execution_date 2021-06-01, it will be running on 2021-07-01 and it will check for the release of 2021-05.
> Alternatively we could set the schedule to '0 0 7 * *' (so it is after the 5th day) and use the 'execution_date' to create the URL. In that case for the execution_date of 2021-06-07, it will be running on 2021-07-07 and it will check for the release of 2021-06. So we would be a bit more up date.
> 
> > If we set the `schedule_interval` to `monthly`, then that will affect 1) the DOI workflow, which runs weekly and has a sensor that waits for the Crossref Metadata DAG to finish. I think that the sensor will wait indefinitely on weeks when the Crossref Metadata DAG does not run. 2) The worker VM is required,
> > Would you be able to find out how you can get the Crossref Metadata sensor in the DOI DAG to only wait on weeks when the Crossref Metadata DAG runs?
> 
> Yeah by default the ExternalTaskSensor expects the schedule_interval+start date to be in sync. We could change this by using the 'execution_date_fn' parameter of the ExternalTaskSensor. I have added a solution in the latest commit.
> What do you think of this solution? I have tested it with this simple DAG:
> 
> ```python
> from datetime import datetime
> 
> from airflow import DAG
> from airflow.operators.dummy_operator import DummyOperator
> from airflow.sensors.external_task_sensor import ExternalTaskSensor
> from observatory.platform.telescopes.telescope import Telescope, Release
> from observatory.platform.utils.telescope_utils import get_prev_execution_date
> 
> 
> class FakeCrossrefTelescope(Telescope):
>     DAG_ID = 'fake_crossref'
>     SCHEDULE_INTERVAL = '@monthly'
> 
>     def __init__(self, dag_id: str = DAG_ID, start_date: datetime = datetime(2020, 6, 7),
>                  schedule_interval: str = SCHEDULE_INTERVAL, catchup=True):
>         """ Construct a Telescope instance.
> 
>         :param dag_id: the id of the DAG.
>         :param start_date: the start date of the DAG.
>         :param schedule_interval: the schedule interval of the DAG.
>         """
>         super().__init__(dag_id, start_date, schedule_interval, catchup=catchup)
> 
>         self.add_setup_task_chain([self.setup_task])
>         self.add_task(self.task)
> 
>     def make_release(self, **kwargs) -> Release:
>         """ Create a single release instance.
> 
>         :param kwargs: the context passed from the PythonOperator. See
>         https://airflow.apache.org/docs/stable/macros-ref.html for more info.
>         :return: A release instance
>         """
>         release_id = f"telescope_{datetime.now()}"
>         release = Release(self.dag_id, release_id)
>         return release
> 
>     def setup_task(self, **kwargs) -> bool:
>         """ This is a custom task that is executed before a release is made. The return value is used for a
>         shortcircuit operator"
> 
>         :param kwargs: the context passed from the PythonOperator. See
>         https://airflow.apache.org/docs/stable/macros-ref.html for more info.
>         :return: True/False
>         """
>         setup = 'successful'
>         return True if setup == 'successful' else False
> 
>     def task(self, release: Release, **kwargs):
>         """ This is a custom task that can use a release instance.
> 
>         :param release: A release instance.
>         :param kwargs: the context passed from the PythonOperator. See
>         https://airflow.apache.org/docs/stable/macros-ref.html for more info.
>         :return: None.
>         """
>         print(f"This is a custom task that uses the release: {release.release_id}")
> 
> 
> telescope = FakeCrossrefTelescope()
> globals()[telescope.dag_id] = telescope.make_dag()
> 
> 
> default_args = {
>     "owner": "airflow",
>     "start_date": datetime(2020, 8, 30)
> }
> SENSOR_ID_CROSSREF_METADATA = 'crossref_metadata_sensor'
> 
> with DAG(dag_id='test_sensor', schedule_interval='@weekly', default_args=default_args, catchup=False) as dag2:
>     # Sensors
>     crossref_metadata_sensor = ExternalTaskSensor(
>         task_id=SENSOR_ID_CROSSREF_METADATA,
>         external_dag_id=FakeCrossrefTelescope.DAG_ID,
>         execution_date_fn=lambda execution_date: get_prev_execution_date(FakeCrossrefTelescope.SCHEDULE_INTERVAL,
>                                                                          execution_date),
>         mode='reschedule')
> 
>     dummy = DummyOperator(
>         task_id='dummy'
>     )
> 
>     crossref_metadata_sensor >> dummy
> ```
> 
> What would be the issue for the worker VM if the schedule interval is not set to '@Weekly'?

That solution looks good! 

In terms of the worker VM, the only thing is that the Crossref Metadata DAG will not run until the worker VM gets turned on. It will be scheduled to run and wait until it turns on. It would always be the Sunday after the 7th of the month. I suppose that would actually work out OK?

aroelo: Thanks, I somehow did not realise this url was available to check the available releases (https://api.crossref.org/snapshots/monthly/2021). I've added some extra logging on the available releases.

I've changed the schedule interval to '0 0 7 * *', so if the worker VM is turned on weekly it will only run when the snapshot of the previous month is already available.

jdddog (commented):
It looks good Aniek, it just needs to be rebased

jdddog (approved):

PR #419, opened 2021-05-18
Update oapen irus uk schema

aroelo:
Update schema to include extra identifiers.
There are some extra identifiers for the book in the output of the 'newer' oapen irus uk data that were not stored in the output file.
The following identifiers are now available:

- proprietary id
- uri (this was previously included as 'item_id')
- doi
- isbn

See https://github.com/The-Academic-Observatory/oapen-irus-uk-cloud-function/pull/3 for corresponding PR of Cloud Function.


jdddog (approved):
Looks good to me

PR #427, opened 2021-06-01
Fix fundref recursion error

aroelo:
Fix error: 'RecursionError: maximum recursion depth exceeded in comparison'

One of the funders was a parent/child of itself, resulting in an infinite recursion loop.
Something similar happened previously and the 'sub_funders' list was added to pick up these cases.
However, this likely only worked for cases where the parent/child directly followed each other, as the sub_funders list was incorrectly set to an empty list for each funder_id.


jdddog (approved):
Nice, thanks Aniek!

PR #428, opened 2021-06-02
Feature/mellon integration

jdddog:
This pull request makes a number of changes to DAGs in order to integrate the various Mellon DAGs together.

JSTOR: to enable JSTOR to whitelist our bot
* Changed random user agent string to the Observatory user agent string.
* Updated Terraform script to make static IPs for VMs.

OAPEN IRUS UK:
* The maximum number of cloud function calls was being reached so I have used a separate variable to set the maximum cloud function instances (rather than max_active_runs) and set the default to unlimited.

OAPEN Metadata:
* dc.date.available: changed to a list field.

ONIX
* Added `date_regex` and `date_format` parameters so that the date from each ONIX file can be extracted for each publisher (each of whom uses a different format).

UCL Discovery:
* Skip if no metadata is available for the period of the release.

ONIX workflow:
* The partners now have partitioned tables rather than sharded tables, so I have enabled it to work with partner data sources that have partitioned tables.


aroelo (approved):
Looks good to me Jamie, thanks for helping clean up/fix some of the Mellon telescopes.
I've just added 2 small commits, fixing some documentation for google analytics and JSTOR.

tuanchien (approved):
Looks ok.

jdddog:
Thanks Aniek and Tuan!

PR #431, opened 2021-06-04
Update oapen irus uk schedule

aroelo:
The data from the previous month was not yet available on the first day of the following month.
I checked today that the data for the month of May is available.

To make sure the data is always available during the DAG run the schedule is changed to run on every 14th day of the month.

@jdddog You initially proposed every second Sunday of the month, but I could not find a way to translate this into a cron expression, what do you think about every 14th day of the month instead?


jdddog (approved):
Yeah I think that should be OK, unless we add a sensor to the onix workflow DAG to wait for the oapen irus uk dag to finish, but we don't need that at the moment.

PR #432, opened 2021-06-08
Reformat schema

aroelo:
Update schema so that country and location info are REPEATED RECORDS containing NULLABLE fields instead of a NULLABLE RECORD containing REPEATED fields. See https://github.com/The-Academic-Observatory/oapen-irus-uk-cloud-function/pull/6 for cloud function PR.


rhosking (approved):
schema changes look good

jdddog (approved):
Looks good to me Aniek

PR #434, opened 2021-06-15
Feature publisher workflow

rhosking:
This work extends the existing onix_workflow. It creates tables in the oaebu dataset and data ready for export to elasticsearch in the oaebu_elastic datasets


tuanchien (commented):
Good job. It's looking good.  Would be good to breakup that massive task creating lots of tables into separate functions for better isolation and testing.

It would also be good to have tests for the tables created in BQ.  I noticed some of the sql templates are quite large.  Not sure how feasible it is to break those up for easier testing, or whether one should just test the final output.

observatory-dags/observatory/dags/workflows/onix_workflow.py
```diff
@@ -580,6 +597,445 @@ def create_oaebu_intermediate_table_tasks(self, data_partners: List[OaebuPartner
...
+                raise AirflowException(
+                    f"create_bigquery_table_from_query failed on {release.project_id}.{output_dataset}.{table_id}"
+                )
+
+    def create_oaebu_output_tasks(self, data_partners: List[OaebuPartners]):
+        """Create tasks for outputing final metrics from our OAEBU data.  It will create output tables in the oaebu dataset.
+        :param oaebu_data: List of oaebu partner data.
+        """
+
+        data_partner_datasets = {
```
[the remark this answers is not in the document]
rhosking: yeah, looks good. Updated that bit

rhosking:
> It would also be good to have tests for the tables created in BQ. I noticed some of the sql templates are quite large. Not sure how feasible it is to break those up for easier testing, or whether one should just test the final output.

Some of those SQL scripts might be able to be improved, but without creating a trail of intermedia tables I'm not sure the logic can useful be broken apart any further. 

Just thinking through how to most usefully test the final actual output now. The workflow itself is better structured now though

observatory-dags/observatory/dags/database/sql/create_book_work_families.sql.jinja2
```diff
@@ -0,0 +1,16 @@
+{# Copyright 2020 Curtin University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
```
jdddog, line 6: This file has no SQL in it, so better to delete it.

observatory-dags/observatory/dags/database/sql/create_book_works.sql.jinja2
```diff
@@ -0,0 +1,16 @@
+{# Copyright 2020 Curtin University
```
jdddog, line 1: This file has no SQL in it, so better to delete it.

jdddog (changes requested):
It looks good Richard, I've just left a few small suggestions for changes.

rhosking:
@tuanchien @jdddog hopefully those changes address all your feedback. Jamie, I also added the skeleton for exporting to Elastic. All the needed variables should be nicely organised, and the tests have been updated to expect that DAG structure. It is just pending any updates/porting you do for those export utilities. It's not super pretty, but it's functionally fine and easier to adapt going forward

tuanchien (dismissed):

jdddog (approved):

PR #436, opened 2021-06-22
Add option to print airflow logging info in test environment

aroelo:
I find it useful to see the airflow task logging when developing/testing a telescope, so I have added a parameter in the `env.create()` method to display the airflow logs.


jdddog (approved):
Looks great Aniek, thanks!

PR #440, opened 2021-06-30
Oaebu elastic import

jdddog:
Make the elastic search import DAG more generic and write unit tests for it.


observatory-dags/observatory/dags/workflows/elastic_import_workflow.py
```diff
@@ -24,37 +24,81 @@
...
+
+
+def load_elastic_mappings_simple(path: str, table_prefix: str):
+    return json.loads(load_file(os.path.join(path, f"{table_prefix.replace('_', '-')}-mappings.json")))
+
+
+def load_elastic_mappings_ao(path: str, table_prefix: str):
+    if not table_prefix.startswith("ao"):
+        raise ValueError("Table must begin with 'ao'")
+    elif table_prefix.startswith("ao_doi"):
```
tuanchien, line 68: what about having an ao_simple prefix for those that want to use load_elastic_mappings_simple
otherwise you will need to keep adding if branches for new workflows

observatory-dags/observatory/dags/workflows/elastic_import_workflow.py
```diff
@@ -24,37 +24,81 @@
...
+        aggregation_level = parts[1]
+        return json.loads(render_template(mappings_path, aggregation_level=aggregation_level))
+
+
+def make_index_prefix(table_id: str):
+    return table_id.replace("_", "-")
+
+
+def make_table_prefix(table_id: str):
+    return table_id[:-8]
```
tuanchien, line 100: add a comment to indicate what the -8 represents

observatory-platform/observatory/platform/elastic/kibana.py
```diff
@@ -234,3 +242,23 @@ def _make_saved_object_url(self, object_type: ObjectType, object_id: str, space_
...
+        return urljoin(self.host, *parts)
+
+    def create_index_pattern(self):
+        pass
+
+    def get_index_pattern(self, index_pattern_id: str, space_id: str = None):
+        url = self._make_index_pattern_url(index_pattern_id, space_id=space_id)
+        response = requests.get(url, headers=self.headers, auth=self.auth)
+
+        if response.status_code == 404:
```
tuanchien, line 260: maybe give a name to 404 for those not familiar with http response codes

aroelo:
@jdddog Is there anything you had in mind for me to look at in particular? :-) 
I'm happy to have a look at the code in general and see if anything stands out to me, but I'm not sure what the code that is added is supposed to do or how/when it is used, so might be hard to give a proper review.

rhosking (changes requested):
I think all the published_year mapping types need to stay/be of the type keyword. If you set them as longs, they get optimised for range queries and some chart types won't be possible (when using a year as a bucket, rather than a continous and they will be displayed with commas, like 2,000 for the year 2000.

However, how I originally had them isn't ideal either. They don't need to be type text as well. so perhaps we can use the multi-field mapping to:

"published_year": {
        "type": "long",
        "fields": {
          "keyword": {
            "type": "keyword",
            "ignore_above": 256
          }
        }

I'm not 100% sure you can use long and keyword together, but I think they are processed independently anyway, so my guess is that it is fine. I saw all the export queries as now has INT64, but I think that should be OK with keywords?

jdddog:
Thanks Richard, I've made those changes and confirmed that you can add a keyword field.

rhosking (approved):
thanks, looks good now!

PR #443, opened 2021-07-13
Remove date check from SQL merge query

aroelo:
Remove usage of date field to check if a matched row is older/newer when merging partitions into a table.

The original SQL
```sql
MERGE
  {{ dataset }}.{{ main_table }} M
USING
  (SELECT {{ merge_condition_field }} AS id, {{ updated_date_field }} AS date FROM {{ dataset }}.{{ partitioned_table }} WHERE _PARTITIONDATE >= '{{ start_date }}' AND _PARTITIONDATE < '{{ end_date }}') P
ON
  M.{{ merge_condition_field }} = P.id
WHEN MATCHED AND M.{{ updated_date_field }} <= P.date OR M.{{ updated_date_field }} is null THEN
  DELETE
```

After removing the date check when a row is matched:
```sql
MERGE
  `{{ dataset }}.{{ main_table }}` M
USING
  (SELECT {{ merge_condition_field }} AS id FROM `{{ dataset }}.{{ partitioned_table }}` WHERE _PARTITIONDATE > '{{ start_date }}' AND _PARTITIONDATE <= '{{ end_date }}') P
ON
  M.{{ merge_condition_field }} = P.id
WHEN MATCHED THEN
  DELETE
```

An issue came up when for a telescope (oapen metadata) the field used for the `updated_date_field` had its mode changed from REQUIRED to REPEATED, because this field can have multiple values.
The `updated_date_field` can't be a REPEATED field in the merge query and there is not really another appropriate field to use instead.

I think this check was a safety check in the first place, because the partition that is merged into the main table should always include newer data. Since this is now causing issues I suggest to just remove this safety check from the merge query completely.

### DIFF
--- a/observatory-dags/observatory/dags/database/sql/merge_delete_matched.sql.jinja2
+++ b/observatory-dags/observatory/dags/database/sql/merge_delete_matched.sql.jinja2
@@ -16,8 +16,8 @@
 MERGE
   `{{ dataset }}.{{ main_table }}` M
 USING
-  (SELECT {{ merge_condition_field }} AS id, {{ updated_date_field }} AS date FROM `{{ dataset }}.{{ partitioned_table }}` WHERE _PARTITIONDATE > '{{ start_date }}' AND _PARTITIONDATE <= '{{ end_date }}') P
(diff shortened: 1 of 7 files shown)


jdddog (approved):
Thanks Aniek, that looks great!

PR #450, opened 2021-07-21
Bugfixes for running workflows in production

jdddog:
This PR fixes a number of issues that were found when running the workflows in production:
* JSTOR telescope: log a warning instead failing when encountering unknown file types.
* Change datetime.datetime types to pendulum.Pendulum: encountered errors in the scheduler because we were using a mix of datetime.datetime and pendulum.Pendulum (timezone aware) types and the scheduler was throwing errors because it can't compare naive datetimes to timezone aware datetimes.
* Add Umich to onix workflow data partners.
* Fix up Kibana index pattern time fields in elastic import.
* Add crossref_events sensor to DOI DAG.

### DIFF
--- a/observatory-dags/observatory/dags/dags/elastic_import.py
+++ b/observatory-dags/observatory/dags/dags/elastic_import.py
@@ -37,10 +37,9 @@ DAG_PREFIX = "elastic_import"
 ELASTIC_MAPPINGS_PATH = make_elastic_mappings_path()
 AO_KIBANA_TIME_FIELDS = [TimeField("^.*$", "published_year")]
 OAEBU_KIBANA_TIME_FIELDS = [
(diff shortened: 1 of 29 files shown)


tuanchien (approved):
Looks ok.

I have a general comment about the pendulum thing.  If we are migrating to Airflow 2, then we need to start using things like pendulum.date and pendulum.datetime.  The latest major branch of pendulum gets rid of the Pendulum class, and the newer airflow has the newer version of pendulum as a dependency.

jdddog:
> Looks ok.
> 
> I have a general comment about the pendulum thing. If we are migrating to Airflow 2, then we need to start using things like pendulum.date and pendulum.datetime. The latest major branch of pendulum gets rid of the Pendulum class, and the newer airflow has the newer version of pendulum as a dependency.

Thanks Tuan, yeah that makes sense

PR #452, opened 2021-07-26
Update ONIX workflow so it always uses latest ONIX release

jdddog:
This PR updates the ONIX Workflow so that it:
* Waits for the ONIX telescope and all data partners to finish via sensors.
* Always uses the latest ONIX release available for an organisation.

The workflow was previously synced with the ONIX telescope and would only generate new data for a release if an ONIX file was processed by the ONIX telescope. This would mean that the ONIX Workflow would not run if there was no new ONIX data, but the data partners had fresh data.


aroelo (approved):
Hey Jamie,

The changes look good to me :-)

PR #461, opened 2021-08-05
Airflow 2 transition

tuanchien:
This branch updates the code base to be compatible with Airflow 2.1.2 (latest stable).  This enables us to access security updates from the latest stable branches.

Major changes:
- Updated requirements.  There might still be undetected conflicts.
- Changed usage of Pendulum to the 2.1.x API.
- Update tests to use create_dag_run() with create() in ObservatoryTestEnvironments
- Use YYYYMMDD string format for release date in xcom messages.  JSON serializable data is preferred over pickle data.
- Use pendulum instead of datetime in the telescopes.  This hasn't been changed for all the utility functions though.
- Change module imports to be consistent with Airflow 2.
- Swap out a few deprecated functions.

Minor changes:
- Update testing documentation.
- Update invocations of airflow cli command to use the new interface.
- Workaround for pbr version conflicts between mag-archiver and observatory platform.

Things requiring more testing as part of PR review:
- End to end functional run to make sure all the DAGs still work as expected.  The tests all pass, but coverage is not 100% for all telescopes.

Airflow 2 introduced a Taskflow API that is a cleaner way to write DAGs. The new way of handling xcoms in Airflow 2 also reduces some of the boiler plate required. No attempt has been made to port our telescope interfaces to use the Taskflow API.  It will be another significant undertaking if we wish to do this since it will require some redesign of all the telescopes.

There are other features in Airflow 2 that have not been configured or enabled.  We can make changes later if we wish to take advantage of them.

Given the auto formating, and having to tweak so many files, the code coverage is likely to have dropped a little bit.  Maybe we can ignore that for this PR.


observatory-platform/observatory/platform/docker/Dockerfile.observatory.jinja2
```diff
@@ -14,7 +14,7 @@
 
 # Author: James Diprose
 
-FROM apache/airflow:1.10.14-python3.7
+FROM apache/airflow:2.1.2-python3.8
```
jdddog, line 17: It would be good if the base Airflow Docker image name / version was a parameter

observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
```diff
@@ -152,24 +152,37 @@ x-build: &build
...
     networks:
       - {{ docker_network_name }}
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 5s
+      timeout: 30s
+      retries: 50
 
   flower:
-    image: apache/airflow:1.10.14-python3.7
```
jdddog, line 163: I realise that the reason this image is used and not the built image is because flower doesn't need the Airflow DAGs to function nor does it need any of the other things we install in the Docker image, so to make it lighter weight the base Airflow image was used.

If the Airflow Docker name / version is set as a parameter somewhere then we could also set this dynamically and use the base Airflow Docker image.

observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
```diff
@@ -249,15 +266,15 @@ services:
...
+    image: "docker.elastic.co/kibana/kibana:7.13.3"
     restart: always
     ports:
       - ${HOST_KIBANA_PORT}:5601
     networks:
       - {{ docker_network_name }}
 
   postgres:
-    image: postgres:12.2
+    image: postgres:13
```
jdddog, line 277: It would be good to stick with postgres:12.2 to avoid problems upgrading existing databases.

jdddog (changes requested):
Hey Tuan, it looks great to me, just a few small comments relating to the Docker and Docker Compose files.

aroelo (approved):

observatory-platform/observatory/platform/docker/Dockerfile.observatory.jinja2
tuanchien: I might do this as part of the work on installation scripts instead since there's already some work done there on templating out versions

observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
tuanchien: Reverting back from build to image.  Will template out as part of installation script work.

observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
tuanchien: reverting back to 12.2.  will template this out as part of installation script work.

jdddog (approved):

PR #462, opened 2021-08-09
Feature/installerscript

tuanchien:
Requires a change to the branch protection policy on github to expect build 3.8, ubuntu 20.04 instead of 3.7, ubuntu 18.04

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
 
       - name: Install packages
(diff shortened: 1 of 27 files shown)


tuanchien:
What does everyone think about there being a 'workflows' folder in the root observatory-platform directory that's created by the installer script, and if people choose to get the oaebu and academic observatory workflows, for those projects to be cloned into the workflows folder?

aroelo:
> What does everyone think about there being a 'workflows' folder in the root observatory-platform directory that's created by the installer script, and if people choose to get the oaebu and academic observatory workflows, for those projects to be cloned into the workflows folder?

I think this setup would make it hard to work on the workflows in these projects, as they are inside the observatory-platform git repository so you won't be able to use git for version control.

PR #463, opened 2021-08-11
Reformat all files using 'black -l120'

aroelo:
Reformat all files in observatory-api, observatory-dags, observatory-platform and tests directories.

Not sure if it should be run on observatory-reports directory.

Currently, for PRs a lot of changes are due to reformatting, making it harder to spot actual differences.
Hopefully running the black formatter on all these files once will make it easier in future PR's to look at changes, because all files are already in the same format.


tuanchien (approved):
Looks good.  Might be easier to just rerun formatting on the current develop again though if there are many conflicts since there might be a fair amount of files changed by the airflow 2 migration.

aroelo:
> Looks good. Might be easier to just rerun formatting on the current develop again though if there are many conflicts since there might be a fair amount of files changed by the airflow 2 migration.

Yeah good point, not sure which one would be merged first.

PR #464, opened 2021-08-12
Reformat all files using 'black -l120'

aroelo:
Reopened version of https://github.com/The-Academic-Observatory/observatory-platform/pull/463 after rebasing and excluding some API files.

Reformat most files in observatory-api, observatory-dags, observatory-platform and tests directories.
Exclude the files in `observatory-api/observatory/api/client directory`, since they are auto generated by OpenAPI Generator

Not sure if it should be run on observatory-reports directory.

Currently, for PRs a lot of changes are due to reformatting, making it harder to spot actual differences.
Hopefully running the black formatter on all these files once will make it easier in future PR's to look at changes, because all files are already in the same format.


jdddog (approved):
Thanks Aniek, that's a great help! I think we can exclude the reports as we will remove it from this repo soon.

PR #469, opened 2021-08-17
Add pre-commit config & shell script

### LINKED ISSUE #465: Usage of pre-commit tool
See https://docs.google.com/document/d/15XhGB74dN_FCBKOmc8aYWR-xGxsqqoXo8nX1vCwQnZ4/ for current workflow of integrating Jira and Github.

The need to mention Jira issue keys in your commit message led me to take a look at different git hooks that can be used before/during your commit.

There is a package [pre-commit](https://pre-commit.com/) that could be used to manage these hooks.
One option would be to add a pre-commit config file that can automatically prefix your commit message with a Jira issue key based on the branch name (see [example](https://github.com/The-Academic-Observatory/jira-integration-test/blob/INF-10/fix_something/.pre-commit-config.yaml)).
There are existing hooks that we could use to do this (https://github.com/avilaton/add-msg-issue-prefix-hook and https://github.com/radix-ai/auto-smart-commit) or add our own simple shell script (trying it out [here](https://github.com/The-Academic-Observatory/jira-integration-test/blob/INF-10/fix_something/prefix_commit_jira.sh))

There are also other existing hooks we could potentially use, for example checking if the files are formatted with 'black'.
A list of supported hooks can be found here: 
https://pre-commit.com/hooks.html

Alternatively we could add a file with this shell script to our repository, with instructions to move this file to `.git/hooks/prepare-commit-msg` if you would like to make use of the git hook.

@rhosking @jdddog @tuanchien Are there any thoughts on using pre-commit, and using other hooks besides the Jira one?

aroelo:
Add pre-commit config & shell script to automate prefixing Jira issue key to commit message, closes #465

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+repos:
+-   repo: local
+    hooks:
+      - id: add-jira-id
+        name: Add Jira ID from branch name
+        description: Finds Jira ID from branch name and adds as prefix to commit message
+        entry: ./bin/prefix_commit_jira.sh
+        language: script
+        stages: [prepare-commit-msg]
--- a/bin/prefix_commit_jira.sh
+++ b/bin/prefix_commit_jira.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# The script below adds the branch name automatically to every one of your commit messages.
+# The regular expression below searches for JIRA issue key's at the start of the branch name.
+# The issue key will be extracted out of your branch name
+
+REGEX_ISSUE_ID="^[A-Z]{1,10}-[0-9]{1,5}"
+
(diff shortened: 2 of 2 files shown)


jdddog (changes requested):
Did these files come from another project and do they have licenses?

aroelo:
> Did these files come from another project and do they have licenses?

No, the config file I created myself and the shell script was inspired by something from stackoverflow (without a license) and made some changes to it as well.

jdddog:
> No, the config file I created myself and the shell script was inspired by something from stackoverflow (without a license) and made some changes to it as well.

Ah ok thanks Aniek, it is probably OK because it is just a small script.

jdddog (approved):

PR #471, opened 2021-08-18
Remove observatory-reports and related files

jdddog:
Moved observatory reports to a separate repo so that it can be archived (perhaps change the name when you archive it): https://github.com/The-Academic-Observatory/observatory-reports


rhosking (changes requested):
Looks good, the only thing I picked up was a header for Reports that was left over in the documentation: https://github.com/The-Academic-Observatory/observatory-platform/blob/feature/remove-reports/docs/index.rst#reports

jdddog:
> Looks good, the only thing I picked up was a header for Reports that was left over in the documentation: https://github.com/The-Academic-Observatory/observatory-platform/blob/feature/remove-reports/docs/index.rst#reports

Thanks, I've removed it

PR #472, opened 2021-08-19
Rename telescope base classes into workflow

jdddog:
This PR:
* Renames the Telescope and AbstractTelescope base classes into Workflow and AbstractWorkflow and associated changes.
* Removed the dependence of the `observatory generate workflow` command on the observatory-dags Python project.


observatory-dags/observatory/dags/dags/elastic_import_telescope.py
```diff
@@ -27,7 +27,7 @@
 )
 from observatory.platform.elastic.elastic import make_elastic_mappings_path
 from observatory.platform.elastic.kibana import TimeField
-from observatory.platform.utils.telescope_utils import make_dag_id
+from observatory.platform.utils.workflow_utils import make_dag_id
```
rhosking, line 30: Does it make sense to call this a telescope? probably need a different postfix

observatory-dags/observatory/dags/workflows/crossref_fundref_telescope.py
```diff
@@ -33,7 +33,7 @@
 from airflow.api.common.experimental.pool import create_pool
 from airflow.exceptions import AirflowException
 from airflow.models.taskinstance import TaskInstance
-from observatory.platform.telescopes.snapshot_telescope import (
+from observatory.platform.workflows.snapshot_workflow import (
```
rhosking, line 36: I still prefer to call it snapshot_telescope, as it is specialised for ingesting data. Would you extend it for another purpose?

observatory-dags/observatory/dags/workflows/crossref_events_telescope.py
```diff
@@ -27,12 +27,12 @@
 import requests
 from airflow.exceptions import AirflowSkipException
 from airflow.models.taskinstance import TaskInstance
-from observatory.platform.telescopes.stream_telescope import (
+from observatory.platform.workflows.stream_workflow import (
```
rhosking, line 30: I would prefer to call it stream_telescope, same reason as the snapshot_telescope

rhosking (changes requested):
I do like using workflow as the base class, as it is very generic and not specialised for ingesting data. But would prefer to keep using telescope for the stream and batch. 

I'm also wondering about creating an analytical_workflow class to extend the base workflow for BigQuery only workflows, possibly with it's own tutorial and command line generation

aroelo:
To me it would make sense to rename the 'Telescope' base class to 'Workflow', but keep the 'StreamTelescope', 'SnapshotTelescope' and 'OrganisationTelescope' as is, like Richard suggested.

This is with the idea in mind that a 'telescope' in general is a specific kind of workflow, namely one that captures a single source of raw data and loads it into BigQuery.

tuanchien:
My 2c is that the telescopes are just workflows as we've set them up. Although one should note that the Release class seems to be quite data ingestion specific.

From an outsider's perspective it's less confusing to call them workflows because people will more readily gauge what it is.  But if the purpose is to build a branding around Telescope as a term that refers to a workflow ingesting datasets, then I guess there's value in keeping that naming.

aroelo (changes requested):
Hey @jdddog, looking good

Just one minor comment, I don't think I'm in favor to rename all the telescope filenames from e.g. `crossref_events` to `crossref_events_telescope`.
I think it makes them a bit long and it would be clearer to keep the workflows/telescopes directories separated.
Would it be possible to keep the `observatory/dags/telescopes` folder and maybe we could rename `observatory/dags/workflows` to `obervatory/dags/aggregation_workflows`?
And perhaps rename `observatory/dags/telescopes` to `observatory/dags/telescope_workflows` as well if we would like to keep the folder names similar in format.

observatory-platform/observatory/platform/cli/generate_command.py
```diff
@@ -73,95 +73,119 @@ def generate_terraform_config(self, config_path: str):
...
+
+        # Make folders
+        for path in [workflow_dst_dir, dags_dst_dir, test_dst_dir, doc_dst_dir]:
+            os.makedirs(path, exist_ok=True)
+
+        # Make init files
+        package_folder = os.path.join(project_path, package_name)
+        tests_package_folder = os.path.join(project_path, "tests", package_name)
+        init_paths = [package_folder, tests_package_folder, workflow_dst_dir, dags_dst_dir, test_dst_dir]
+        for path in init_paths:
```
tuanchien, line 137: create only if it doesn't exist

observatory-platform/observatory/platform/cli/cli.py
```diff
@@ -222,17 +222,26 @@ def generate():
...
+@click.argument("package_name", type=str)
+@click.argument("workflow_type", type=click.Choice(["Workflow", "StreamTelescope", "SnapshotTelescope"]))
+@click.argument("workflow_name", type=str)
+def workflow(project_path: str, package_name: str, workflow_type: str, workflow_name: str):
+    """Generate boiler plate code for a new workflow.
+
+    project_path: The Python project path.
+    package_name: The Python package name.
+    workflow_type: Type of workflow. Options are Workflow, StreamTelescope, SnapshotTelescope.
+    workflow_name: Name of your new workflow.
```
tuanchien, line 237: Might be good to give an example of usage, and the summary of the directory structure it produces.

tuanchien (changes requested):

observatory-platform/observatory/platform/cli/generate_command.py
aroelo: This will be covered by the changes in the telescope documentation branch, at least that is what I'm assuming, because there are still quite a lot of difference in the `generate workflow` command for these two branches.

observatory-platform/observatory/platform/cli/cli.py
aroelo: This will be covered by the changes in the telescope documentation branch, at least that is what I'm assuming, because there are still quite a lot of difference in the `generate workflow` command for these two branches.

jdddog:
> Hey @jdddog, looking good
> 
> Just one minor comment, I don't think I'm in favor to rename all the telescope filenames from e.g. `crossref_events` to `crossref_events_telescope`.
> I think it makes them a bit long and it would be clearer to keep the workflows/telescopes directories separated.
> Would it be possible to keep the `observatory/dags/telescopes` folder and maybe we could rename `observatory/dags/workflows` to `obervatory/dags/aggregation_workflows`?
> And perhaps rename `observatory/dags/telescopes` to `observatory/dags/telescope_workflows` as well if we would like to keep the folder names similar in format.

Yeah the file names are a bit longer, however they are also clearer because they have the full name of the class inside them. There will also be about half of the files in each `dags` and `workflows` module because all of the workflows will be split into two different projects, `academic-observatory-workflows` and `oaebu-workflows`.

The workflows all have to be in the same dags module, because only one dags folder is loaded into Airflow for each workflows project. I figure that since all of the files have to be together in the same dags folder, the workflow classes can all be together in their own workflows module to be consistent.

rhosking:
> because only one dags folder is loaded into Airflow for each workflows project. I figure that since all of the files have to be together in the same dags folder, the workflow classes can all be together in their own workflows module to be consistent.

Do we still need to split the DAG and the workflow/telescope file into two parts anymore? most of the DAG files have become quite minimalistic, and the remainder could probably be simplified through some improvements to the Platform API?

jdddog:
> > because only one dags folder is loaded into Airflow for each workflows project. I figure that since all of the files have to be together in the same dags folder, the workflow classes can all be together in their own workflows module to be consistent.
> 
> Do we still need to split the DAG and the workflow/telescope file into two parts anymore? most of the DAG files have become quite minimalistic, and the remainder could probably be simplified through some improvements to the Platform API?
> 
> There could still perhaps be a folder with classes/utilities that are specific to Mellon/Academic-Observatory that don't make sense to stay in the observatory-platform perhaps

Yes, we still need to separate the `dags` folder from the `workflows` folder.
* dags folder: this determines what Airflow dags are created and loaded in the Airflow instance.
* workflows folder: this is where the workflow classes live.

Some workflows will be used in multiple projects, for instance, the ElasticImportWorkflow, will be used in both the academic-observatory-workflows and the oaebu-workflows. Having a separate `dags` and `workflows` folder means that we can control which workflows actually get instantiated for a workflows project.

It also makes it simpler from the perspective of the workflow code generation because the folders that code is generated into are always the same, i.e. the `dags` and `workflows` folders.

observatory-platform/observatory/platform/cli/cli.py
jdddog: I've added a bit more help text, I think project path and package name will disappear after your changes Aniek:
```
Usage: observatory generate workflow [OPTIONS] PROJECT_PATH PACKAGE_NAME [Work
                                     flow|StreamTelescope|SnapshotTelescope]
                                     WORKFLOW_NAME

  Generate a new workflow.

  PROJECT_PATH is the Python project path.
  PACKAGE_NAME is the Python package name.
  WORKFLOW_TYPE is the type of workflow.
  WORKFLOW_NAME is the the name of your new workflow.

  For example, the command: observatory generate workflow /path/to/my-workflows-project my_workflows_project Workflow MyWorkflow

  Will generate the following files and folders:

  └── my-workflows-project
      ├── docs
      │   ├── index.rst
      │   └── my_workflow.md
      ├── my_workflows_project
      │   ├── dags
      │   │   ├── __init__.py
      │   │   └── my_workflow.py
      │   ├── __init__.py
      │   └── workflows
      │       ├── __init__.py
      │       └── my_workflow.py
      └── tests
          └── my_workflows_project
              ├── __init__.py
              └── workflows
                  ├── __init__.py
                  └── test_my_workflow.py

Options:
  --help  Show this message and exit.

```

To generate that example folder structure, you can run the generate command and then install the program `tree` and then run tree from the folder above `my-workflows-project` and copy paste.

aroelo: Yeah do you mean assuming that the command is run from inside a project?

Also, is there a need to have the package_name as a subdirectory in the tests directory?
Could it be `tests/workflows/test_my_workflow.py`?

aroelo (commented):
Hi @jdddog, nice work on adding the schema_path as a variable everywhere, making it possible to have different paths for different modules :-)

The only things is I wonder if we could change it to `schema_folder` or `schema_folder_path`? 
I think it is a bit confusing now in the workflow files, where it looks like `schema_path` is the path to the schema file, instead of the folder.

jdddog:
> Hi @jdddog, nice work on adding the schema_path as a variable everywhere, making it possible to have different paths for different modules :-)
> 
> The only things is I wonder if we could change it to `schema_folder` or `schema_folder_path`?
> I think it is a bit confusing now in the workflow files, where it looks like `schema_path` is the path to the schema file, instead of the folder.

Thanks Aniek, yeah sure I can do that.

jdddog:
> Hi @jdddog, nice work on adding the schema_path as a variable everywhere, making it possible to have different paths for different modules :-)
> 
> The only things is I wonder if we could change it to `schema_folder` or `schema_folder_path`?
> I think it is a bit confusing now in the workflow files, where it looks like `schema_path` is the path to the schema file, instead of the folder.

That's all sorted now.

rhosking (approved):

PR #473, opened 2021-08-20
Upgrade Terraform 0.13 -> 1.0.5

aroelo:


jdddog (commented):
Thanks Aniek, it looks good to me. I've tested upgrading the observatory-develop environment from 0.13 to 1.05 and it has gone smoothly.

It would just be good to update the Terraform download links in the documentation so that they refer to the latest versions (the Mac and Linux instructions).

The Linux one should be:
```bash
sudo curl -L "https://releases.hashicorp.com/terraform/1.0.5/terraform_1.0.5_linux_amd64.zip" -o /usr/local/bin/terraform
```

PR #474, opened 2021-08-23
Create Google Cloud Secrets for all Airflow Variables

aroelo:
While working on the telescope documentation I was documenting why we use a custom 'AirflowVariable' class to get values for the Airflow variables.
The explanation is below:

> #### Secrets backend issue, use custom AirflowVariable class
> In the cloud environment that is deployed with terraform, Airflow uses Google Cloud Secret Manager as a secrets
>  backend and both the Airflow variable and connections are stored in there as secrets.
> 
> Unfortunately, there is currently an issue with Airflow's method to obtain variables when using the secrets backend.  
> This issue has been resolved by implementing a custom `AirflowVariable` class inside 
> `observatory-platform/observatory/platform/utils/airflow_utils.py`.  
> This class should always be used to get variables instead of the standard Airflow `Variable` class inside `airflow
> .models.variable.py`.
> 
> For example, to get a variable:  
> ```python
> from observatory.platform.utils.airflow_utils import AirflowVariable, AirflowVars
> variable = AirflowVariable.get(AirflowVars.DOWNLOAD_BUCKET)
> ```
> 
> The problem occurs when Airflow tries to get a variable while the Google Cloud Secrets Manager is set as a secrets
>  backend, but the variable is not stored as a Google Cloud Secret.  
> This is the case for some Airflow variables that are often used in the telescopes (test_data_path, data_path, 
> download_bucket, transform_bucket).  
> They are not set in the 'airflow_variables' section of the `config-terraform.yaml` file and this means that these
>  variables are not stored as Google Cloud Secrets.
> They do exist as valid Airflow variables, but as environment variables instead of Google Cloud Secrets.  
> The search order for variables/connections for Airflow is not configurable and with a secrets backend enabled the
>  order is:   
> ```
> secrets backend > environment variables > metastore
> ```
> 
> The issue is that the current Airflow method to get variables from the secrets backend will return an error when a
>  Google Cloud Secret can not be found.
> This error prevents Airflow from searching for a variable in remaining places (environment variables & metastore),
>  so the value for a variable can never be resolved if it does not exist as a Google Cloud Secret.   
> The custom `AirflowVariable` class solves this by catching the error in a Try/Except clause, meaning that Airflow
>  will continue to look for a variable in the remaining places.

While I still think there is a flaw in Airflow's method to just raise an error whenever a secret does not exist, a better workaround for us might be to just make sure that all the Airflow Variables listed in the AirflowVars class exist (ofcourse the user defined variables depend on whether they are added to the config file).

I've added the `data_path`, `download_bucket` and `transform_bucket` variables as these were the only ones missing.
We can then remove the custom Airflow class and the relevant part from the documentation.

UPDATE:
To prevent excessive costs for accessing secrets I customised the `workflow_path` function to first look for an environment variable instead of a google cloud secret.

I also found that the updated secret_manager_client has a fix so that when a secret cannot be found this is caught with an exception, instead of raising an error. This means that the custom AirflowVariable class is not needed anymore for this reason alone.

### DIFF
--- a/observatory-platform/observatory/platform/observatory_config.py
+++ b/observatory-platform/observatory/platform/observatory_config.py
@@ -30,7 +30,7 @@ from cerberus import Validator
 from cryptography.fernet import Fernet
 
 from observatory.platform.terraform_api import TerraformVariable
-from observatory.platform.utils.airflow_utils import AirflowVariable, AirflowVars
+from observatory.platform.utils.airflow_utils import AirflowVars
 from observatory.platform.utils.config_utils import module_file_path
 from observatory.platform.utils.config_utils import observatory_home as default_observatory_home
 from observatory.platform.utils.jinja2_utils import render_template
--- a/observatory-platform/observatory/platform/observatory_config.py
+++ b/observatory-platform/observatory/platform/observatory_config.py
@@ -896,9 +896,7 @@ class TerraformConfig(ObservatoryConfig):
         """
 
         # Create airflow variables from fixed config file values
-        data_path = "/opt/observatory/data"  # Hardcoded in docker-compose.observatory.yml.jinja2
-        variables = [AirflowVariable(AirflowVars.DATA_PATH, data_path),
-                     AirflowVariable(AirflowVars.ENVIRONMENT, self.backend.environment.value)]
+        variables = [AirflowVariable(AirflowVars.ENVIRONMENT, self.backend.environment.value)]
 
         if self.google_cloud.project_id is not None:
             variables.append(AirflowVariable(AirflowVars.PROJECT_ID, self.google_cloud.project_id))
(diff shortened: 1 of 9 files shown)


observatory-platform/observatory/platform/observatory_config.py
```diff
@@ -896,10 +896,14 @@ def make_airflow_variables(self) -> List[AirflowVariable]:
         """
 
         # Create airflow variables from fixed config file values
-        variables = [AirflowVariable(AirflowVars.ENVIRONMENT, self.backend.environment.value)]
+        data_path = "/opt/observatory/data"  # Hardcoded in docker-compose.observatory.yml.jinja2
```
jdddog, line 899: I think you could not add `data_path` as a cloud secret because it is always set from an environment variable (as hardcoded in the Docker Compose file.

You could also remove these two lines from the Docker Compose file now: https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2#L89-L90

jdddog (commented):
It looks good to me Aniek. Just a few small changes like what we have discussed on Slack

jdddog (approved):

PR #475, opened 2021-08-24
Remove observatory-dags

jdddog:
This PR removes the observatory-dags project from the observatory platform.


observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
```diff
@@ -285,7 +286,7 @@ services:
     networks:
       - {{ docker_network_name }}
     healthcheck:
-      test: ["CMD", "pg_isready", "-U", "airflow"]
+      test: ["CMD", "pg_isready", "-U", "{{ postgres_user }}"]
```
rhosking, line 289: I never really got to the bottom of my problems, I think they were largely around docker not having enough memory (ES is a memory hog FYI). I actually changed it back to just being airflow and nothing broke. So not sure exactly what is going on

rhosking: It's worth considering making ES+Kibana being an optional flag in the observatory start command (maybe that is already the case, I should double check). Elasticsearch want's to grow to around a 4gb memory footprint, and if you are testing components that don't rely on it, it would be helpful to not turn that on.

rhosking (changes requested):
Looks good, lots of deleted files to scroll through! just a couple of small points, might be better to address them with future changes

docs/conf.py
```diff
@@ -19,7 +19,6 @@
 
 from recommonmark.transform import AutoStructify
 
-from generate_schema_csv import generate_csv, generate_latest_files
```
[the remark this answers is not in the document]
aroelo: I would be in favour to keep the function in each project repo, so you can run `make html` independent of the observatory-platform.
I think the idea is that the workflows project can be completely independent right?
I did add the code to the observatory platform, and it is then copied into the docs repository of a project when you generate a new project using CLI: 
https://github.com/The-Academic-Observatory/observatory-platform/tree/telescope_documentation/observatory-platform/observatory/platform/cli/templates/generate_project

observatory-platform/requirements.txt
```diff
@@ -1,39 +1,55 @@
+# Airflow
 apache-airflow==2.1.3
 apache-airflow-providers-slack==4.0.0
 apache-airflow-providers-http==2.0.0
```
aroelo, line 4: Nice work organising all of these in different sections :-)

aroelo (approved):
Looks good to me!

observatory-platform/requirements.txt
jdddog: Haha thanks, I wanted to figure out what requirements could be dropped and thought it would make it easier in the future

docs/conf.py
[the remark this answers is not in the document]
jdddog: If we want that `generate_schema_csv` function to be in the platform, then we need to build the whole Observatory Platform Python project and install all of it's dependencies just to build the docs, which will slow down building the docs.

Ideally this function would be a Sphinx plugin and in it's own package, that should be future work though.

observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
jdddog: Yeah we can do that in the future. Could you add a ticket for it on JIRA?

jdddog:
> Looks good to me!
> 
> EDIT: @jdddog Not sure if this is the right place for a comment, but some thoughts about the new project structure for the workflows.
> For the `observatory generate project` command I have this structure right now:
> 
> ```
>   For example, the command: observatory generate project /path/to/my-workflows-project my_workflows_project
> 
>   Will generate the following files and folders:
> 
>   └── my-workflows-project
>       ├── docs
>       │   ├── _build
>       │   ├── _static
>       │   ├── _templates
>       │   ├── workflows
>       │   ├── conf.py
>       │   ├── generate_schema_csv.py
>       │   ├── index.rst
>       │   ├── make.bat
>       │   ├── Makefile
>       │   └── requirements.txt
>       ├── my_workflows_project
>       │   ├── __init__.py
>       │   ├── dags
>       │   │   └── __init__.py
>       │   ├── database
>       │   │   ├── __init__.py
>       │   │   └── schema
>       │   │       └── __init__.py
>       │   ├── utils
>       │   │   └── __init__.py
>       │   └── workflows
>       │       └── __init__.py
>       └── tests
>           ├── __init__.py
>           ├── setup.cfg
>           ├── setup.py
>           └── workflows
>               └── __init__.py
> ```
> 
> I saw that you have the tests for the workflows inside `my-workflows-project/my_workflows_project/workflows/tests` instead.
> I don't mind changing the structure for the generate command, but just curious in general what the approach is for how to organise the unit tests?
> If there are any utils functions for example, where would these tests go?
> Directly in `my-workflows-project/my_workflows_project/tests`?
> Or `my-workflows-project/my_workflows_projects/tests/utils`?
> Or if it would be similar to the workflows, in `my-workflows-project/my_workflows_project/utils/tests`

Thanks Aniek.

So for your example with a utils sub package: `my-workflows-project/my_workflows_project/utils`
The tests would go here: `my-workflows-project/my_workflows_project/utils/tests`

> 
> And a different comment, shouldn't the observatory-platform be listed as a requirement in here https://github.com/The-Academic-Observatory/academic-observatory-workflows/blob/develop/requirements.txt?

Yes it should, I haven't done that yet though, I'll do that once I've sorted out enabling the observatory-platform to run when it has been installed from a source distribution and when those builds are published on PyPI.

observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
rhosking: yeah, will do

rhosking (approved):

PR #485, opened 2021-09-01
Remove TelescopeTypes class and usage from observatory-api module

aroelo:
The init_db() function that is used to seed a database in a new session should not use defined oaebu TelescopeTypes.
The TelescopeTypes class has been removed and is added in the oaebu workflows project (https://github.com/The-Academic-Observatory/oaebu-workflows/pull/9)

I thought it would make sense to keep the create_session and init_db functionality as it is, but instead pass an optional list of telescope types that will be used to seed the database.


jdddog (changes requested):
Thanks Aniek, I think it would be better to just remove the seed database code as `create_session` is never going to be called with `telescope_types` being set.

We need to think through in more detail how to seed the database.

aroelo:
> Thanks Aniek, I think it would be better to just remove the seed database code as `create_session` is never going to be called with `telescope_types` being set.
> 
> We need to think through in more detail how to seed the database.

Ah sure, I can remove the code. I thought maybe it would be called in `app.py`, possibly with other things as well than TelescopeTypes, if we end up using the API for other metadata.
But we could work on this when reformatting the API in general.

jdddog (approved):

PR #487, opened 2021-09-02
Fix flower and database initialisation errors

jdddog:
This PR fixes a problem with the flower container not starting and the database not initialising on a clean deploy with Terraform.

The flower container would not start because the GoogleCloudSecrets backend wasn't installed in the flower container.


aroelo (approved):
Thanks for the fix!

PR #493, opened 2021-09-12
Inf 39/port unpaywall to snapshot telescope

tuanchien:
This PR is required for the unpaywall telescope port.

Testing:
- Add simple threaded httpserver for testing use

Utilities
- Add get_observatory_http_header to create simple header dict using custom user agent
- Add get_fiename_from_url to get a filename from a http url
- Add get_chunks function to split lists into constant size (unless last chunk) chunks.
- Add get_airflow_connection_url to pull a url from an airflow connection, validate it, and add trailing "/" if necessary.
- Add converter function for csv to jsonl files.
- Add http get response functions for simple interfaces to standardise getting http raw text response, xml -> dict, json->dict.
- Add AsyncHttpFileDownloader with download_file and download_files interfaces for downloading files using http. Allows custom headers to be used in http connection.
  - download_files allows concurrent downloading through asyncio and aiohttp. Supports retry on failure with exponential backoff.
  - download_file piggybacks off download_files.  No speed benefit from asyncio, but provides a simpler interface.

Snapshot telescope:
- Add upload_downloaded as a snapshot telescope task.  I noticed a lot of the upload_downloaded tasks in snapshot telescopes are identical in implementation. They all just upload the download_files list of files from the release object to the download_bucket in the cloud.  Since this is a standard pattern we have adopted, it may as well just be part of the snapshot telescope implementation.
- Add download, extract, transform tasks to template.

Stream telescope:
- Add download, upload_downloaded, extract, transform tasks to template.


observatory-platform/observatory/platform/utils/workflow_utils.py
```diff
@@ -1320,3 +1308,78 @@ def make_table_name(
...
+    :param chunk_size: Size of chunks.
+    :return: The next chunk from the input list.
+    """
+
+    n = len(input_list)
+    for i in range(0, n, chunk_size):
+        yield input_list[i : i + chunk_size]
+
+
+class AsyncHttpFileDownloader:
```
jdddog, line 1326: At the moment we are using third party code to download files, specifically the function `get_file` in `data_utils.py`.
* https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory-platform/observatory/platform/utils/data_utils.py
* https://github.com/The-Academic-Observatory/observatory-platform/blob/develop/observatory-platform/observatory/platform/utils/progbar_utils.py
* https://github.com/The-Academic-Observatory/observatory-platform/blob/87a78df26a44f0d890a87fc2ba50e3ba1fd653fc/observatory-platform/observatory/platform/utils/file_utils.py#L91-L111: this function should be removed as well, it isn't from our project and should be in data_utils.py if it is going to stay.
* https://github.com/The-Academic-Observatory/observatory-platform/blob/87a78df26a44f0d890a87fc2ba50e3ba1fd653fc/observatory-platform/observatory/platform/utils/file_utils.py#L59-L88: this function should be removed as well, it isn't from our project and should be in data_utils.py if it is going to stay.

We could almost completely replace it with your code, but there is one thing missing, which is the ability to verify if a downloaded file has a particular file hash (you'll see the parameters `file_hash` and `hash_algorithm` it in `get_file`), which is important to make sure that a partly downloaded file is not accepted as a fully downloaded file after a task crashes and restarts. 

Would you be able to add `file_hash=None` as an optional parameter?
* We could just use md5 for now.
* If `file_hash` hash is supplied, then validate the file.
  * If the validation fails, then re-download the file.
  * If the validation succeeds, skip downloading the file. 
* If no `file_hash` is supplied:
  * Is the default behaviour to overwrite existing files?

We would also then need to replace instances of `get_file` throughout the code.

This would also mean that we could remove the following dependencies from requirements.txt:
```
# To remove in future
numpy>=1.20.3,<2
six>=1.16.0,<2
```

observatory-platform/observatory/platform/utils/json_util.py
```diff
@@ -35,3 +36,23 @@ def to_json_lines(items: List[dict], serialize_custom_types_func=None) -> str:
     json_lines = "".join(json_records)
 
     return json_lines
+
+
+def convert_csv_to_jsonlines_file(*, csv_file: str, jsonl_file: str):
```
jdddog, line 41:
```suggestion
def csv_to_jsonlines(*, csv_file: str, jsonl_file: str):
```

observatory-platform/requirements.txt
```diff
@@ -34,13 +34,19 @@ json_lines>=0.5.0,<1 # Reading, including jsonl.gz
...
+aiohttp
+
 
 # SFTP
 pysftp>=0.2.9<1
 paramiko>=2.7.2,<3
 
 # Utils
 natsort>=7.1.1,<8
+backoff==1.11.*
```
jdddog, line 46: Could you put a looser definition for the backoff version? e.g. `backoff>=1,<2`, since the observatory platform can be depended on by downstream projects, it shouldn't be too strict about library versions.

observatory-platform/requirements.txt
```diff
@@ -34,13 +34,19 @@ json_lines>=0.5.0,<1 # Reading, including jsonl.gz
 # HTTP requests and URL cleaning
 requests>=2.25.0,<3
 tldextract>=3.1.1
+aiohttp
```
jdddog, line 37: Perhaps put a version information here

jdddog (changes requested):
Hey Tuan, it looks good to me, just some comments about the file downloading. 

Also, you might be able to give the file downloader a simpler name, e.g. FileDownloader, or put it into it's own module, e.g. file_downloader.py and remove the class, since the class doesn't do much.

tuanchien:
- Added a rewrite of file hash validation methods in preparation of removing the third party code.
- Shorted the csv to json converter function name as per request.
- Requirements for backoff loosened.  Requirement for aiohttp tightened.

AsyncHttpFileDownloader changes:
 - `download_file`, `download_files` are functions in the module `observatory.platform.utils.http_download`
 - Extended to use a worker pool instead of chunks
 - The downloader skips a file if the file exists on the filesystem and its hash matches a supplied hash.
 - The downloader will requeue a download once if a downloaded file does not have a hash matching a supplied hash.

I had a look at the code in the repos using get_file and the current third party hash functions _hash_file, validate_file, and it's in quite a few places.  It's better to create a separate ticket to do that removal. I will do this in JIRA soon.

jdddog (approved):
Looks good to me

PR #494, opened 2021-09-13
Bugfix deployment

jdddog:
This PR addresses the following:
* Docker Compose / Airflow 2:
  * Received the error "daemonic processes are not allowed to have children" when tasks ran that use multiprocessing.Pool, to address it added `AIRFLOW__CORE__EXECUTE_TASKS_NEW_PYTHON_INTERPRETER ` to the Docker Compose file. This is the same error described in [this Stack Overflow post](https://stackoverflow.com/questions/68878031/is-multiprocessing-pool-not-allowed-in-airflow-task-assertionerror-daemonic).
  * Set Docker Compose volume paths correctly for editable workflows packages when deployed to Terraform.
* Terraform:
  * Update TerraformBuilder so that it builds with the latest changes. 
  * Observatory API: `postgres` connection prefix deprecated in PostgresSQL 1.4, so changed in Terraform file to `postgresql`.
* Address inconsistent use of dates:
  * Change type hints pendulum.datetime to pendulum.DateTime (the class, not function).
  * Change datetime.datetime calls to pendulum.datetime.
  * Make `select_table_shard_dates` return List[pendulum.Date]
  * Add a `make_release_date` function, which returns a pendulum.DateTime instance, which is required for some of the downstream functions that use it.
* load_dags.py:
  * When DagBag has import errors, raise an exception that has a message with all of the errors so that the Dag import errors are visible in the Airflow UI.
  * Moved some of the functions in `load_dags.py` into workflow_utils so they could be used elsewhere.
* Added missing `data_files` in config.cfg.


rhosking (approved):
cool - tests are passing now, so looks like sensible changes

PR #497, opened 2021-09-15
Add black to precommit config

aroelo:
I also ran it on all files using `pre-commit run --all-files`


jdddog (approved):
Thanks Aniek, that's great

PR #498, opened 2021-09-19
Inf 40/create unpaywall data feed telescope

tuanchien:
- bugfix: get_airflow_connection_url
- add get_airflow_connection_password
- add unzip_files function
- add find_replace_file  (sed cli replacement)
- add fn to wrap shell cmd calls. treats non zero exit as error.

Using unzip/find_replace makes the platform slightly more portable compared to calling a cli program in bash. unzip is faster than cli + bash pipe equivalent.  find_replace is about same speed as sed + bash pipe.

### DIFF
--- a/observatory-platform/observatory/platform/utils/file_utils.py
+++ b/observatory-platform/observatory/platform/utils/file_utils.py
@@ -333,3 +333,45 @@ def zip_files(file_list: List[str]):
                 shutil.copyfileobj(f_in, f_out)
 
     return zip_list
+
+
+def unzip_files(*, file_list: List[str], output_dir: str = None):
+    """Gunzip the list of files.
+
+    :param file_list: List of files to unzip.
+    :param output_dir: Optional output directory.
+    """
+
+    for file_path in file_list:
+        if file_path[-3:] != ".gz":  # Skip files without .gz extension
+            continue
+
+        logging.info(f"Unzipping file {file_path}")
+
+        if output_dir is None:
+            output_dir = os.path.dirname(file_path)
+
+        filename = os.path.basename(file_path)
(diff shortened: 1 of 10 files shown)


observatory-platform/observatory/platform/utils/file_utils.py
```diff
@@ -333,3 +333,45 @@ def zip_files(file_list: List[str]):
                 shutil.copyfileobj(f_in, f_out)
 
     return zip_list
+
+
+def unzip_files(*, file_list: List[str], output_dir: str = None):
```
jdddog, line 338: Perhaps give this function a name that communicates that it is unzipping gzip files, e.g. gunzip_files

observatory-platform/observatory/platform/utils/proc_utils.py
```diff
@@ -52,3 +56,21 @@ def stream_process(proc: Popen, debug: bool) -> Tuple[str, str]:
...
+
+
+def run_bash_cmd(cmd: str):
+    """Run a command in the bash shell and wait until it's done.  Log the output.
+    Raise an exception where there's stderr.
+
+    :param cmd: Command to run.
+    """
+
+    p = Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, executable="/bin/bash")
```
jdddog, line 68: Maybe add `shell=True` as a keyword argument in run_bash_cmd so that it can be turned off if required.

jdddog (changes requested):
Looks good Tuan, just a few small comments.

tuanchien:
Changed:
- unzip_files to gunzip_files
- added run_cmd with shell/executable parameters if you just want to run a generic command

jdddog (approved):

PR #503, opened 2021-09-23
Inf 40/unpaywall data feed fixes

tuanchien:


jdddog (commented):
It looks good to me Tuan, I just wonder if the download_info Dict should be a dataclass to make the interface clearer.

tuanchien:
- DownloadInfo added.
- Validation check added in function to make sure we're getting strings or DownloadInfo
- Add prefix_dir parameter to allow prefixing the filename paths (helpful to control save location if using http headers to do filename discovery).

jdddog (approved):

PR #508, opened 2021-10-01
Fix StreamTelescope BQ merge functions

aroelo:
- Use task instance start date to determine days between last merge and now, instead of release end date. The release end date is still used to determine the ingestion partition date.
- Check if transform files are available when creating bq_info in case batch_load is set to true. 
Because the bq_delete_old, bq_append_new and cleanup tasks have a trigger rule of 'none_failed', these tasks won't be skipped even when the setup tasks return False (because e.g. no new releases are available). For the case of the unpaywall feed it would try to load a non-existing partition into the main table.

### DIFF
--- a/observatory-platform/observatory/platform/workflows/stream_telescope.py
+++ b/observatory-platform/observatory/platform/workflows/stream_telescope.py
@@ -268,11 +268,11 @@ class StreamTelescope(Workflow):
 
         dag_run: DagRun = kwargs["dag_run"]
         ti: TaskInstance = kwargs["ti"]
-        start_date = pendulum.instance(get_prev_start_date_success_task(dag_run, ti.task_id))
-        end_date = release.end_date
-        if (end_date - start_date).days + 1 >= self.bq_merge_days:
+        start_date = pendulum.instance(get_prev_start_date_success_task(dag_run, ti.task_id)).start_of("day")
+        now = pendulum.instance(ti.start_date).start_of("day")
+        if (now - start_date).days >= self.bq_merge_days:
             logging.info(
-                f"Deleting old data from main table using partitions after {start_date} and on or before" f" {end_date}"
+                f"Deleting old data from main table using partitions after {start_date} and on or before {release.end_date}"
             )
             bq_load_info = get_bq_load_info(
                 self.dag_id, release.transform_folder, release.transform_files, self.batch_load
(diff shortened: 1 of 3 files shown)


jdddog:
Is this one ready for review?

aroelo:
> Is this one ready for review?

Yes, thanks for asking :-)

observatory-platform/observatory/platform/workflows/stream_telescope.py
```diff
@@ -268,19 +268,19 @@ def bq_delete_old(self, release: StreamRelease, **kwargs):
 
         dag_run: DagRun = kwargs["dag_run"]
         ti: TaskInstance = kwargs["ti"]
-        start_date = pendulum.instance(get_prev_start_date_success_task(dag_run, ti.task_id))
-        end_date = release.end_date
-        if (end_date - start_date).days + 1 >= self.bq_merge_days:
+        start_date = pendulum.instance(get_prev_start_date_success_task(dag_run, ti.task_id)).start_of("day")
```
jdddog, line 271: Should `start_date =` be using the start date of the previous task or the `release.end_date` of the previous release that was merged? Since the partition date used by `bq_load_ingestion_partition` is `release.end_date`?

Also, in the SQL queries `insert_from_partitions` and `merge_delete_matched`, `_PARTITIONDATE > '{{ start_date }}' AND _PARTITIONDATE <= '{{ end_date }}'` will always skip the start date? 

And the start_date supplied to the query will always skip partitions because the partition is set to `release.end_date`, which always occurs before the date that the previous task ran.

I wonder if it would make it more robust to query the BigQuery table to find out the last partition date? It is probably the best source of truth for that information in this case because the BigQuery table is what is being updated.

observatory-platform/observatory/platform/workflows/stream_telescope.py
```diff
@@ -268,19 +268,19 @@ def bq_delete_old(self, release: StreamRelease, **kwargs):
...
+        now = pendulum.instance(ti.start_date).start_of("day")
+        if (now - start_date).days >= self.bq_merge_days:
```
jdddog, line 273: I personally think it would make things a lot simpler and easier to maintain to remove the merge days feature.

jdddog (commented):
Hey Aniek, I've left a few comments about the wider stream telescope and some questions about the dates that are being used - it could just be my lack of understanding about how some of the BigQuery SQL scripts work.

aroelo:
As discussed this morning on our team standup it will be better for now to remove the bq_merge functionality. The costs of running multiple merge queries might be slightly higher, but it will make overall maintenance easier.
I will close this PR and open a new one to remove the functionality.

PR #509, opened 2021-10-03
Inf 32/installer script after repo separation

tuanchien:


observatory-platform/observatory/platform/observatory_config.py
```diff
@@ -676,8 +780,12 @@ def __init__(
...
         """
 
-        self.backend = backend
-        self.observatory = observatory
+        self.backend = backend if backend is not None else Backend()
+        self.observatory = (
+            observatory
+            if observatory is not None
+            else Observatory(package="observatory-platform", package_type="editable")
+        )
```
aroelo, line 788: I think you could just set these default values directly in the init method:
```python
backend: Backend = Backend(),
observatory: Observatory = Observatory(package="observatory-platform", package_type="editable"),
```

observatory-platform/observatory/platform/observatory_config.py
```diff
@@ -906,6 +984,151 @@ def load(cls, path: str):
...
+        with open(path, "w") as f:
+            self.save_backend(f)
+            self.save_observatory(f)
+            self.save_terraform(f)
+            self.save_google_cloud(f)
+            self.save_airflow_connections(f)
+            self.save_airflow_variables(f)
+            self.save_workflows_projects(f)
+
+    def save_backend(self, f: io.IOBase):
```
aroelo, line 1014: For this case and all other uses of 'f' in this document, the typing of 'f' should be `TextIO` instead, this can be imported with `from typing import TextIO`. See also https://stackoverflow.com/a/61000390

observatory-platform/observatory/platform/cli/generate_command.py
```diff
@@ -221,6 +255,38 @@ def generate_workflow(self, project_path: str, package_name: str, workflow_type:
                 f.write(f'    {workflow_module} = "{workflow_module}"\n')
             print(f"- Updated the identifiers file: {identifiers_dst_file}")
 
+    def generate_local_config_interactive(self, config_path: str, *, workflows: Set[str], oapi: bool):
```
aroelo, line 258: Should workflows be a 'List[str]' instead? From the usage/tests it seems that it a list is passed on.

observatory-platform/observatory/platform/cli/generate_command.py
```diff
@@ -221,6 +255,38 @@ def generate_workflow(self, project_path: str, package_name: str, workflow_type:
...
+
+        file_type = "Observatory Config"
+        click.echo(f"Generating {file_type}...")
+
+        config = InteractiveConfigBuilder.build(backend_type=BackendType.local, workflows=workflows, oapi=oapi)
+
+        config.save(config_path)
+        click.echo(f'{file_type} saved to: "{config_path}"')
+
+    def generate_terraform_config_interactive(self, config_path: str, *, workflows: Set[str], oapi: bool):
```
aroelo, line 274: Should workflows be a 'List[str]' instead? From the usage/tests it seems that it a list is passed on.

tests/observatory/platform/cli/test_generate_command.py
```diff
@@ -241,3 +251,576 @@ def test_write_rendered_template(self, mock_click_confirm):
...
+        expected = Backend(type=BackendType.terraform, environment=Environment.staging)
+        InteractiveConfigBuilder.config_backend(config=config, backend_type=expected.type)
+        self.assertEqual(config.backend.type, expected.type)
+        self.assertEqual(config.backend.environment, expected.environment)
+
+    @patch("observatory.platform.cli.generate_command.click.prompt")
+    def test_config_observatory_filled_keys(self, m_prompt):
+        observatory = Observatory(
+            airflow_fernet_key="IWt5jFGSw2MD1shTdwzLPTFO16G8iEAU3A6mGo_vJTY=",
+            airflow_secret_key=("a" * 16,),
```
aroelo, line 372:
```suggestion
            airflow_secret_key=("a" * 16),
```
With the comma included it generates a tuple instead of a string.

aroelo (changes requested):
Hey @tuanchien, I've tried the install command as well as generating the config file interactively.

For the `pip` way of installing, is this just to use the platform rather than contributing/editing anything in the platform? Maybe this should be a bit more explicit in the documentation.
I tried to start the platform after installing this way, but it was just stuck at the 'building..' step for at least ~15 min. It might be something specific going wrong on my laptop rather than the instructions, I could look into it a bit more. Did this work for you?
```
Observatory Platform: all dependencies found                                    
  Docker:
   - path: /usr/local/bin/docker
   - running
  Docker Compose:
   - path: /Users/284060a/PycharmProjects/test_install/venv/bin/docker-compose
  config.yaml:
   - path: ./config.yaml
   - file valid
  Host machine settings:
   - observatory home: /Users/284060a/.observatory
   - host-uid: 1001
Observatory Platform: building...    
```

I then tried to install using the `git` way and these are the things I noticed while going through it:  
- When using the `install_observatory_platform` command it would be useful if a question is prompted for the config path. Now it asks if I want to overwrite `/Users/284060a/.observatory/config.yaml` which already exists. I don't, but I would like to create the config file automatically, at a different location. Same for the terraform configuration.
- There is an error when trying to install Docker, I think you can just use `brew install` instead of `brew cask install`
```
-----------------
Installing Docker
-----------------
Error: Unknown command: cask
```
- It would also be great if the config path mentioned above could be passed on to `configure_observatory_platform`
- It would be helpful if the virtual environment that is created during `install_observatory_platform` is also activated during the install process, or alternatively to make a comment in the documentation with a reminder to activate the virtual environment.
- During the `configure_observatory_platform` 'local' script
   - This question is asked: 
   - > What type of observatory platform installation did you perform? A git clone is an editable type, and a pip install is a pypi type. (editable, sdist, pypi) [editable]: editable
   
   How would I get the sdist type?
   -  When entering an observatory config directory, it would be great if it would make the dir if it does not exist yet, instead of giving an error.
   - This question is a little confusing, it says that when using the installer script the workflows projects are automatically added to the config file, so I confirmed "N", but I don't see the workflows projects in the config file. Should I still say "Y"?:
   - > Configuring workflows projects. If you opted to install some workflows projects through the installer script then they will be automatically added to the config file for you. If not, e.g., if you installed via pip, you will need to add those projects manually now (or later).
Do you want to add workflows projects? [y/N]: 
    -  I think this would be better off as a y/N question like all of the other ones:
    - > Do you wish to enable ElasticSearch and Kibana? [True]:  
    - Only oaebu workflows is installed, because the git clone url is pointing to oaebu-workflows for the academic-observatory-workflows project. Fixing this might fix this error as well:
` ERROR: workflows/academic-observatory-workflows[tests] is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).`
    -  If installing the workflows projects inside the observatory-platform project, will it still be possible to commit changes to the repositories of those workflows projects? See also https://github.com/The-Academic-Observatory/observatory-platform/pull/462#issuecomment-913969270 Perhaps there is a way to do this, I'm just not sure about it.
    - After running the script with default values I get the following error:
```
Observatory Platform: dependencies missing                                      
  Docker:
   - path: /usr/local/bin/docker
   - running
  Docker Compose:
   - path: /Users/284060a/PycharmProjects/test_install_git/observatory-platform/venv/bin/docker-compose
  config.yaml:
   - path: /Users/284060a/PycharmProjects/academic_observatory/observatory-platform/config_v2.yaml
   - file invalid
    - observatory.airflow_secret_key: is not a valid secret key. Reason: Key b'0c0fdc27feae75fd6e88c01c157dacc1043591422924c7c739bc55681aff' is not base64.
 ```
- During the `configure_observatory_platform` 'terraform' script
    -  It would be great if the `Configuring settings for the main VM that   runs the Airflow scheduler and webserver` bit could stand out a bit more with bold/italic/bullet point. I kind of missed this line and was confused which VM I was configuring.
    - I wouldn't refer to the `weekly on-demand VM` using the term `weekly`, as this is determined by the settings of the vm_create workflow. So it might be monthly or fortnightly with different settings. I think `worker VM` and `main VM` would be good descriptions as this is what we generally use throughout the code/documentation.
    - Maybe you could add default values for these questions, the same as they are in the `config-terraform.yaml.jinja2` file:
```
Google CloudSQL db tier: db-custom-2-7680
Google CloudSQL backup start time, e.g., 13:00: 23:00
Configuring settings for the main VM that   runs the Airflow scheduler and webserver
Machine type, e.g., n2-standard-2: n2-standard-2
Disk size (GB), e.g., 50: 50
```
-  When using the `configure_observatory_platform` script for terraform, these key/values are stored in the config file under the 'observatory' section, but should not be included as they are not used for terraform:
```
  observatory_home: /Users/284060a/.observatorytestinstall
  redis_port: 6379
  flower_ui_port: 5555
  airflow_ui_port: 8080
  elastic_port: 9200
  kibana_port: 5601
  docker_network_name: observatory-network
  docker_network_is_external: False
  docker_compose_project_name: observatory
  enable_elk: True
  api_package: observatory-api
  api_package_type: pypi
```

aroelo:
Hey @tuanchien,

Nice work on all those changes. I'm just trying out the scripts again with the new changes.
I only have two small notes:  

- When getting the `install.sh` using curl it actually immediately starts executing the script, so it is a bit out out of line with the documentation saying I have to make it executable and then execute it.
- When using 'source' to install, the virutalenv is installed in the directory from which the install.sh script is run, not in the directory of the cloned repository. E.g. the installer at the end says:
> You can start the observatory platform after the restart by first activating the Python virtual environment with:
  source /Users/284060a/PycharmProjects/test_install/observatory-platform/observatory_venv/bin/activate

But the virtualenv is actually at `/Users/284060a/PycharmProjects/test_install/observatory_venv/bin/activate`

Note: I was writing this comment and testing out the installation when @jdddog merged the branch, not sure what the best way is to notify others that I'm currently reviewing?

PR #510, opened 2021-10-04
Fix double bucket delete race

tuanchien:
calling env.create() more than once on the same ObservatoryEnvironment object can lead it the context trying to clean up the same bucket more than once, causing sporadic test failures.


jdddog (approved):
Thanks Tuan, looks good to me

PR #520, opened 2021-10-31
remove unused wos/scopus code

tuanchien:
Remove:
- legacy code in observatory platform no longer used by the wos/scopus telescopes after porting.
- unused code written for wos/scopus that was never deployed.


jdddog (approved):
Looks good Tuan

PR #521, opened 2021-10-31
installer: add https/ssh clone option

tuanchien:
- Add question for selecting cloning method for git (ssh or https).
- Use apt to install virtualenv initially. This mitigates an installation issue from a completely fresh Ubuntu install.
- Change formatting of options and default option in questions.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -58,21 +58,39 @@ function check_system() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


jdddog (approved):
Thanks Tuan, looks good.

PR #524, opened 2021-11-09
Enable a bucket path to specified in azure_to_google_cloud_storage_transfer

jdddog:
Enable a sink bucket path to specified in azure_to_google_cloud_storage_transfer, for compatibility with the MAG changes.


rhosking (approved):
looks good

PR #526, opened 2021-11-14
Add Dataset, DatasetRelease, DatasetStorage API extension

tuanchien:


jdddog (commented):
Hey Tuan, the code looks great. 

It would be good to have a short use case based tutorial for the new API features (just for the Python client library). I was initially thinking that this could be done as another PR, but now I think that it would be useful to evaluate and give suggestions on the design of the API.

It can be a really simple page in the docs:

## Contents
* Use Case 1 Title
* Use Case 2 Title

## Use Case 1 Title
Description of use case 1.
```python
import abc
```

## Use Case 2 Title
Description of use case 2.
```python
import abc
```

I will add to this page after you are done, for the existing features of the API.

tuanchien:
- Added documentation for getting, creating, modifying, deleting Dataset, DatasetRelease, DatasetStorage objects using the observatory platform Python API.
- Made a simplification to the get_dataset_release API.

jdddog (approved):
Thanks Tuan, looks good

PR #528, opened 2021-11-18
Inf 278/workflow xcom cleanup

tuanchien:
- Update importlib-metadata to 3.7.3 [Fixes recent general unit test failures in the observatory-platform]
- Add code to cleanup tasks in the telescope templates to delete old (>= 31 days old) xcom messages.


jdddog (approved):
Thanks Tuan, it looks good to me.

PR #532, opened 2021-12-14
Fix VM warning message on slack

aroelo:
The 'Airflow' channel on slack has been receiving warning messages about how long the VM has been on every time the dag runs.
Instead, it should only send a message every X no. of hours. The last time a warning message was sent should be retrieved from an xcom object.
Currently, the xcom object with this information is not properly retrieved. The xcom is pushed with the task id 'check_dags_status', but the workflow attempts to pull the xcom object with the non-existing task id 'check_runtime_vm'.

I have updated the task id in the relevant workflow and also updated some of the unit tests.


tuanchien (approved):

jdddog (approved):
Thanks Aniek

PR #537, opened 2022-01-06
BigQuery bytes processed

jdddog:
Update the BigQuery bytes processed API so that:
* Set BigQueryBytesProcessed.total as a BIGINT. The number of bytes can easily be larger than a 32 bit integer.
* Instead of updating a single record for each day, which could cause a race condition when multiple queries run at once, a new record is added for each query that is run (bigquery_bytes_processed/post). 
* bigquery_bytes_processed/get now returns the aggregated bytes used over the last 24 hours. 
* Instead of allowing the user to set the date of the query, it is automatically set on the server (the created parameter).


observatory-api/observatory/api/server/openapi.yaml.jinja2
```diff
@@ -191,6 +191,7 @@ paths:
           schema:
             $ref: '#/definitions/QueryResponse'
         401:
+          {% if api_client -%}description: API key is missing or invalid{%- endif %}
```
aroelo, line 194: Does the api_client not pick up the description from the UnauthorizedError reference defined below?

aroelo (commented):
Hey @jdddog, looks good to me.
Did you re-run the openapi generator?
It looks like some of the manual documentation from Tuan disappeared in the files in docs/api/*.md, not sure if you have rewritten/moved all of these to the new file `docs/api/api_use_cases.md` or if this is unrelated.

aroelo (approved):

observatory-api/observatory/api/server/openapi.yaml.jinja2
jdddog: The API client and Cloud Endpoints pick it up fine without the description field, but sphinxcontrib.openapi, which generates the REST API documentation, throws an error if the description field is missing. The OpenAPI specification used in the documentation is generated when api_client is True.

This sphinxcontrib.openapi PR fixes it, however it has been pending since Oct 2021: https://github.com/sphinx-contrib/openapi/pull/120

aroelo: Mmm I see, that's a bit annoying, hopefully it will get updated soon :-)

aroelo (approved):
Looks good @jdddog

PR #538, opened 2022-01-10
Add functionality to use multiple instances for Elastic Import workflow

aroelo:
As discussed on slack and thanks to your pointers @jdddog:
- Update Elastic Import workflow to use an Airflow connection based on a variable defined in ElasticImportConfig. This makes it possible to use multiple Airflow connections for different Elasticsearch/Kibana instances.
- Update the Elastic and Kibana clients to use an API key instead of username/password.
- Update unit tests to use ElasticEnvironment instead of deployed elastic cluster.

The config.yml file needs to be updated to use the following Airflow connections:
```
elastic_main: https://<api_key_id>:<api_key>@<elastic_address>:<port>
elastic_public: https://<api_key_id>:<api_key>@<elastic_address>:<port>
kibana_main: https://<api_key_id>:<api_key>@<kibana_address>:<port>
kibana_public: https://<api_key_id>:<api_key>@<kibana_address>:<port>
```

The PR's for academic-observatory-workflows and oaebu-workflows will have to be reviewed and merged after this one as well.

EDIT: the github secrets do not have to be updated anymore since all unit tests now use the ElasticEnvironment

### DIFF
--- a/docs/workflows/elastic_import.md
+++ b/docs/workflows/elastic_import.md
@@ -4,7 +4,7 @@
 +------------------------------+---------+
 | Summary                      |         |
 +==============================+=========+
-| Runs on remote worker        | True    |
+| Runs on remote worker        | False   |
 +------------------------------+---------+
 | Catchup missed runs          | False   |
 +------------------------------+---------+
@@ -34,8 +34,8 @@ Elasticsearch
   Privileges: 'all'
   
 Kibana  
-- To create/delete index patterns, for a limited set of kibana spaces
-  Spaces: <required spaces>
+- To create/delete index patterns, for a limited set of kibana spaces  
+  Spaces: <required spaces>  
   Privileges: Customize -> Management -> Index Pattern Management: All
 
 ### Creating an API key
--- a/docs/workflows/elastic_import.md
+++ b/docs/workflows/elastic_import.md
@@ -1,0 +1,109 @@
+# Elastic Import
+
+```eval_rst
++------------------------------+---------+
+| Summary                      |         |
++==============================+=========+
+| Runs on remote worker        | True    |
++------------------------------+---------+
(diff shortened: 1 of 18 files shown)


docs/workflows/elastic_import.md
```diff
@@ -0,0 +1,109 @@
+# Elastic Import
+
+```eval_rst
++------------------------------+---------+
+| Summary                      |         |
++==============================+=========+
+| Runs on remote worker        | True    |
```
jdddog, line 7: It runs on the main VM so this should be False.

jdddog (commented):
It looks good to me Aniek, thanks for adding the documentation! Just a comment to change the documentation to say that it doesn't run on the remote VM.

docs/workflows/elastic_import.md
aroelo: Ah yeah, thanks for picking that up

observatory-platform/observatory/platform/elastic/kibana.py
```diff
@@ -75,23 +56,36 @@ class ObjectType(Enum):
 class Kibana:
     HTTP_NOT_FOUND = 404
 
-    headers = {"Content-Type": "application/json", "kbn-xsrf": "true"}
+    headers: dict = None
```
jdddog, line 59: This should be a parameter in the constructor

jdddog (changes requested):
Hey Aniek, it looks good to me, just a minor comment.

observatory-platform/observatory/platform/elastic/kibana.py
aroelo: Thanks, just changed this

jdddog (approved):
Thanks Aniek, looks good to me.

PR #539, opened 2022-01-21
Parameterise select table shard date limit

tuanchien:


jdddog (approved):
Thanks for the fix Tuan!

PR #541, opened 2022-02-01
Update API key documentation and fix AIRFLOW__CORE__SQL_ALCHEMY_CONN

jdddog:
Changed:
* postgres+psycopg2 to postgresql+psycopg2 as the first option does not work anymore with SQL Alchemy.
* Updated the Elastic and Kibana API key documentation to show you how to create an API key for Airflow using the Kibana Dev Console. I personally prefer to create an API key in this way, because then you don't need to create and manage another user account and password.


aroelo (commented):
Hey @jdddog, I agree it is nice that you don't have to manage another user/role by creating the API key directly, but would it be possible to view the assigned permissions to the key after it has been created?

I can't find it anywhere in the Elastic documentation and it would be harder to manage or update these permissions if we don't know the current permissions.

Alternatively, we could record the current permissions somewhere internally, but that means managing it there in another place so not sure if it is still advantageous compared to creating a user/role.

PR #542, opened 2022-02-07
OpenAlex telescope changes

aroelo:
This PR includes changes required for the OpenAlex telescope:

- Include possibility to use transfer manifest file in `aws_to_google_cloud_storage_transfer` function
- Add test method to assert the number of bytes in a table
- Move the `get_bq_load_info` function from utils to inside the StreamTelescope class, so it can be overridden for specific use cases


jdddog (approved):
Thanks Aniek, it looks good to me.

PR #554, opened 2022-05-06
Fix/api update

jdddog:
Changes to enable deploying and testing:
* Added OpenAlex dataset types.
* Added WorkflowTypeId class.
* Fixes for the local API server:
  * make_observatory_api: when password not set in Airflow Connection then don't raise error and instead do not set api_key in Configuration.
  * Refactor the Docker related code so that the editable version of the API package is installed rather than the PyPI version. This will need some changes later.


tuanchien:
Maybe change the default port to 5002 isntead of 5001 in this branch too so it doesn't conflict with the tests (which use a mixture of 5000 and 5001).

jdddog:
> Maybe change the default port to 5002 isntead of 5001 in this branch too so it doesn't conflict with the tests (which use a mixture of 5000 and 5001).

Thanks Tuan, I've changed that.

tuanchien (approved):

PR #557, opened 2022-05-16
streamtelescope: add diff merge by lex order

tuanchien:
Add stream telescope task that can merge multiple updates into a single update.
It applies updates files in lexicographic order.

### DIFF
--- a/observatory-platform/observatory/platform/workflows/stream_telescope.py
+++ b/observatory-platform/observatory/platform/workflows/stream_telescope.py
@@ -16,8 +16,11 @@
 
 import logging
 from typing import Dict, Tuple, List
-
 import pendulum
+import os
+import json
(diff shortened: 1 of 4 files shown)


jdddog (approved):
Thanks Tuan, it looks good to me and has fixed the bug with Unpaywall.

aroelo:
@jdddog @tuanchien A bit late to the party, but I'm wondering if this task would fit better as just a part of the Unpaywall telescope? Or are there use cases for this task to be used by other 'stream telescopes'? It seems quite specific for unpaywall.

tuanchien:
In Unpaywall it acts as a substitute for using the Airflow catchup mechanic. In general it addresses the case when you want to update a main table with a series of individual update files which need to be applied in a specified order.

PR #558, opened 2022-05-17
Build Observatory API Image & Push to Google Cloud Artifact Registry

jdddog:
To make system deployment easier, I'm finishing the automatic image building, beginning with automatically building the Observatory API image and providing the ability to customise which images are used for the Observatory API backend and the Cloud Endpoints Runtime.

The main changes in this PR:
* Added the Github Action, Build Observatory API. This uses the Github Actions docker actions to build and push the Docker images to the Google Cloud Artifiact Registry, which has superseded the Google Cloud Container registry. The images are pushed into the three Google Cloud projects (develop, staging and production) so that Cloud Run can pull the images from the same Google Cloud project it is running from (easier to automate permissions).
* In TerraformConfig Api class, added the parameters `api_image` and `er_image`. Since you can now specify the full path to the Docker images, it means that Terraform doesn't need to automatically figure out when to update the image. It also means that we can upgrade the Cloud Endpoints Runtime image without modifying the Terraform files. Examples of the full paths to the Docker images:
  * api_image: `us-docker.pkg.dev/your-project-name/observatory-platform/observatory-api:0.3.1`
  * er_image: `gcr.io/endpoints-release/endpoints-runtime-serverless:2.36.0`

Smaller changes:
* Added a separate requirements.txt file for when the Observatory API is built for Cloud Run, so that it doesn't interfere with the Airflow installation. It means for instance, that we can add the markupsafe==2.0.1 fix without needing to upgrade Airflow yet.

The Build Observatory API Github Action should run on any push (branch or tag), so if you make changes in a branch and want to test that version of the API you can.

I've tested the changes on Terraform.


aroelo (approved):
Thanks @jdddog, looks good!

PR #559, opened 2022-05-25
Add functionality to use BigQuery snapshots

aroelo:
* Add utility function to create a BigQuery snapshot
* Add utility function to delete older rows when there are multiple rows with the same identifier. The older rows are based on the partition date.
* Add method/task to stream telescope to create a BQ snapshot

### DIFF
--- a/observatory-platform/observatory/platform/utils/gc_utils.py
+++ b/observatory-platform/observatory/platform/utils/gc_utils.py
@@ -32,7 +32,7 @@ import pendulum
 from airflow.models import Variable
 from google.api_core.exceptions import BadRequest, Conflict
 from google.cloud import bigquery, storage
-from google.cloud.bigquery import LoadJob, LoadJobConfig, QueryJob, SourceFormat
+from google.cloud.bigquery import LoadJob, LoadJobConfig, QueryJob, SourceFormat, CopyJobConfig, CopyJob
 from google.cloud.bigquery.job import QueryJobConfig
 from google.cloud.exceptions import Conflict, NotFound
 from google.cloud.storage import Blob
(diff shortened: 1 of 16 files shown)


observatory-platform/observatory/platform/workflows/stream_telescope.py
```diff
@@ -333,14 +326,28 @@ def bq_append_new(self, release: StreamRelease, **kwargs):
...
+
+        :param release: a StreamRelease instance
+        :param kwargs: The context passed from the PythonOperator.
+        :return: None.
+        """
+        logging.info("Creating a BigQuery snapshot")
+        bq_load_info = self.get_bq_load_info(release)
+        for _, main_table_id, _ in bq_load_info:
+            destination_table_id = bigquery_sharded_table_id(main_table_id + "_snapshots", release.end_date)
+            create_bigquery_snapshot(self.dataset_id, main_table_id, self.dataset_id, destination_table_id)
```
jdddog, line 340: Perhaps not for this PR as it will affect many other workflows, but it would be cleaner to explicitly pass in a project id.

jdddog (commented):
Thanks Aniek, it looks good. Just some small comments, not necessarily for this PR as they might have wider consequences. Also, it will need to be rebased.

observatory-platform/observatory/platform/utils/gc_utils.py
```diff
@@ -291,6 +291,49 @@ def create_empty_bigquery_table(
     return table
 
 
+def create_bigquery_snapshot(
+    source_dataset_id: str,
+    source_table_name: str,
+    destination_dataset_id: str,
+    destination_table_name: str,
+    project_id: str = None,
```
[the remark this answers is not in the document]
aroelo: Ah okay, yeah I can definitely change it for this function, but good point to have that as a general status quo.

observatory-platform/observatory/platform/workflows/stream_telescope.py
```diff
@@ -31,31 +33,22 @@
...
     def __init__(
-        self,
-        dag_id: str,
-        start_date: pendulum.DateTime,
-        end_date: pendulum.DateTime,
-        first_release: bool,
-        download_files_regex: str = None,
-        extract_files_regex: str = None,
-        transform_files_regex: str = None,
+            self,
```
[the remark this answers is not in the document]
aroelo: Mmm I think I accidentally didn't use Black on some of these committed files. I tried working with multiple windows in Pycharm, but it seems that if you do that you can't use git hooks?
I reverted back to just using a single window for each repository, but may have missed formatting on some files. I will run it on the changed files and see if that fixes it.

observatory-platform/observatory/platform/workflows/templates/workflow_stream.py.jinja2
```diff
@@ -47,13 +47,14 @@ class {{ workflow_class }}(StreamTelescope):
         dataset_id: str = "your_dataset_id",
         dataset_description: str = "The your_dataset_name dataset: https://dataseturl",
         merge_partition_field: str = "id",
+        workflow_id: int = 0,
```
jdddog, line 50: Could we make this a positional argument rather than a keyword argument? Since you have to specify it every time you instantiate a StreamTelescope subclass.

jdddog (changes requested):
It looks good to me Aniek, just a small change, other than that and rebasing it looks good to go.

observatory-platform/observatory/platform/workflows/templates/workflow_stream.py.jinja2
aroelo: Yeah sure, you mean just in this jinja2 template? It is already a positional argument in the actual StreamTelescope class. This jinja2 template is for an example workflow that uses the StreamTelescope. For the actual existing telescopes that use the StreamTelescope the workflow_id is set in the DAG file, so this template is an exception I guess.

jdddog: Yeah in the jinja2 template and the other StreamTelescope subclasses, but I'm not sure if it will cause a lot of changes to the tests for the other StreamTelescope subclasses?

aroelo: I will have a look :-)

jdddog (approved):
Thanks Aniek, it looks good to me.

PR #560, opened 2022-05-26
Separate config loading from config use

jdddog:
This PR separates the loading of configuration files from the classes and methods that use them, which I'm doing in preparation of automatically building images of the system. This means that the CLI can handle config file loading, validation, and logging validation errors with a single function called 'load config.' Duplicate config file loading code in PlatformCommand and TerraformCommand can be removed. I've also added more tests and moved several files to more logical locations.


aroelo (approved):
Thanks @jdddog, it looks much better organised!

PR #561, opened 2022-06-01
Config default update

keegansmith21:
Updated the creation of the default config.yaml file. It is now clear that 'package' refers to a path and not the literal package name


jdddog (commented):
Thanks for your PR Keegan, by default the package is the name of the package on PyPI, so by default it is set to `observatory-platform` https://pypi.org/project/observatory-platform/. It can also be the name of the package and the version e.g. `observatory-platform==0.4.0`.

When `package_type` is editable, then package is the path to the local project cloned from Github.

Perhaps the documentation for this needs to be updated somewhere so that it is clearer?

keegansmith21:
@jdddog ready for re-review!

jdddog (approved):
Thanks for the update to the command line tool documentation Keegan!

PR #563, opened 2022-06-09
Create directories that are mounted as volumes with Docker

aroelo:
@keegansmith21 and I had some issues when starting the observatory platform and syncing the files in the 'data' directory generated by the workflows from the container to the host.

It turned out that the 'data' directory on the host did not exist yet, so these files can then not be synced. I think it would be good to make sure that all directories that are mounted as volumes exist before starting the observatory platform.
As far as I can tell these are only the 'data', 'logs' and 'postgres' directories.


jdddog (approved):
Thanks Aniek, the changes look good to me.

PR #566, opened 2022-06-13
Raise exception if dataset missing when adding releases

tuanchien:


jdddog (changes requested):
Thanks Tuan, it looks good, it just needs to be rebased as it has the changes from the tags branch and now develop has been updated.

jdddog (approved):

PR #569, opened 2022-06-14
Use find_free_port in more unit tests

tuanchien:


jdddog (approved):
Thanks Tuan, it looks good to me.

PR #571, opened 2022-06-17
Propagate dag tag to telescopes

tuanchien:


jdddog (approved):
Thanks Tuan, looks good.

PR #572, opened 2022-06-23
Inf 418/add dag tags to workflows

tuanchien:
Setting up platform changes for adding vm create/destroy and elastic import workflow tagging.

Ran into an issue when testing locally. Error was similar to:
https://github.com/apache/airflow/issues/22289

which seems to be fixed in Airflow 2.3.0

However, erasing my local postgres db and starting the platform up with clean db did not present the problem.


jdddog (approved):
Thanks Tuan!

PR #573, opened 2022-06-23
Fix seed_db service in docker-compose

aroelo:
Move seed_db service inside if statement, only create when the backend type is set to 'local'


jdddog (approved):
Thanks Aniek, I'll figure out how to make the seeding work on a remote environment in a different PR.

PR #574, opened 2022-07-04
api: cascade delete for Workflow and Dataset

tuanchien:
Instead of refusing to delete Workflow and Dataset objects with dependents, delete the dependent objects as well.

This requires a database change to be performed if upgrading an existing db:

alter table dataset 
drop constraint dataset_workflow_id_fkey,
add constraint dataset_workflow_id_fkey foreign key (workflow_id) references workflow(id) on delete cascade;

alter table dataset_release
drop constraint dataset_release_dataset_id_fkey,
add constraint dataset_release_dataset_id_fkey foreign key (dataset_id) references dataset(id) on delete cascade;


jdddog:
Closed by https://github.com/The-Academic-Observatory/observatory-platform/pull/606

PR #575, opened 2022-08-01
jsonl comparison fix

keegansmith21:
Currently, a json file can be compared to a bigquery table. An issue arises because bigquery will pull some fields as objects if done through the Python client vs exporting the table through the bigquery console as json. This PR resolves the issue by dumping both jsons to strings and then reloading them before making assertions.

This blocks The-Academic-Observatory/oaebu-workflows#102

### DIFF
--- a/observatory-platform/observatory/platform/utils/test_utils.py
+++ b/observatory-platform/observatory/platform/utils/test_utils.py
@@ -69,6 +69,7 @@ import threading
 import time
 import unittest
 import uuid
+import json
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from http.server import SimpleHTTPRequestHandler, ThreadingHTTPServer
@@ -653,13 +654,13 @@ class ObservatoryTestCase(unittest.TestCase):
         actual_content = None
         try:
             rows = self.bigquery_client.list_rows(table_id)
-            actual_content = [dict(row) for row in rows]
+            actual_content = json.loads(json.dumps([dict(row) for row in rows], default=str))
         except NotFound:
             pass
 
         self.assertIsNotNone(rows)
         if expected_content is not None:
-            for row in expected_content:
+            for row in json.loads(json.dumps(expected_content, default=str)):
                 self.assertIn(row, actual_content)
                 actual_content.remove(row)
             self.assertListEqual(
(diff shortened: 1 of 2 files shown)


observatory-platform/observatory/platform/utils/test_utils.py
```diff
@@ -653,13 +654,13 @@ def assert_table_content(self, table_id: str, expected_content: List[dict] = Non
...
             rows = self.bigquery_client.list_rows(table_id)
-            actual_content = [dict(row) for row in rows]
+            actual_content = json.loads(json.dumps([dict(row) for row in rows], default=str))
         except NotFound:
             pass
 
         self.assertIsNotNone(rows)
         if expected_content is not None:
-            for row in expected_content:
+            for row in json.loads(json.dumps(expected_content, default=str)):
```
tuanchien, line 663: Will it still work without this change, given we're expecting `expected_content: List[dict]` ?

tuanchien (commented):
Looks good Keegan. I just left a question about one of the line changes.

observatory-platform/observatory/platform/utils/test_utils.py
keegansmith21: Yes, and also with the use-case that I've been testing with (loading a .json from file). The function only requires a List[dict] though, so I figured that it's possible it's handed a table in the same format that it loads the ```actual_content```
```
rows = self.bigquery_client.list_rows(table_id)
actual_content = [dict(row) for row in rows]
```
In this case, the ```actual_content``` would be stringified, but the ```expected_content``` would not be

jdddog: I think the problem is that `dict()` doesn't convert objects to strings and the dates end up as a `FakeDate` object from the BigQuery (not sure why it isn't a datetime), whereas the dates loaded from JSON are strings.

jdddog, line 663: We shouldn't need to convert the expected items if we have loaded them from JSON or JSON Lines.

jdddog (changes requested):

observatory-platform/observatory/platform/utils/test_utils.py
keegansmith21: No problem, I've reverted the expected content in https://github.com/The-Academic-Observatory/observatory-platform/pull/575/commits/b8414f278133ecae1b0a1a8ca8d2f29ad225cdf8

jdddog (approved):
Thanks Keegan, it looks good to me.

PR #577, opened 2022-08-09
Updated docker compose process to pipe its stderr to stdout

keegansmith21:
Docker compose subprocess appears to hang on occasion. Part of the reason is that the stdout stream of compose is never filled and its stderr is never printed. This update should fix the printing issue. I haven't had `observatory platform start` hang since implementing the change, but i'll wait until i'm more confident it isn't still broken before requesting a merge.

### DIFF
--- a/observatory-platform/observatory/platform/docker/compose_runner.py
+++ b/observatory-platform/observatory/platform/docker/compose_runner.py
@@ -121,7 +121,7 @@ class ComposeRunner(Builder):
         proc: Popen = subprocess.Popen(
             self.COMPOSE_ARGS_PREFIX + [self.compose_file_name] + args,
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
             env=env,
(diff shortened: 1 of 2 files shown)


keegansmith21:
After some testing and a bit more research i'm satisfied that this fixes the issue that me and @alexmassen-hane were having.

The reason for the hanging process is due to a limitation of python's subprocess streaming to stdout/stderr. If the stream fills with enough data, the process will [deadlock](https://docs.python.org/3/library/subprocess.html#:~:text=retry%20the%20wait.-,Note,-This%20will%20deadlock).

This wouldn't have been an issue if stream_process worked as intended as it would be continuously flushing the stream. However, the function only flushed stdout in the expected way, before dumping all of stderr once the process is finished. Since the process locks and never finishes (because docker seems to write to stderr only), the process never writes anything or finishes.

I fixed this by piping the stderr docker process output to stdout.

jdddog (approved):
Thanks for fixing the bug Keegan, it looks good to me.

PR #578, opened 2022-08-19
Added prefix functionality to ObservatoryEnvironment class

alexmassen-hane:
When unit tests run and are stopped part way through, they can leave behind unwanted buckets and datasets in Google Cloud and BigQuery. 

Functions have been added to remove unnecessary old test buckets and datasets that can be run at the start of each test that are older than 7 days from the Google Storage and BigQuery that share a common prefix. Each prefix is unique to the test and organisation name and will not remove buckets or datasets that are being used by other tests. 

Each unit test still needs to be modified to add this pre-run clean up step.

### DIFF
--- a/observatory-platform/observatory/platform/utils/gc_utils.py
+++ b/observatory-platform/observatory/platform/utils/gc_utils.py
@@ -34,10 +34,10 @@ import pendulum
 from airflow.models import Variable
 from google.api_core.exceptions import BadRequest, Conflict
 from google.cloud import bigquery, storage
-from google.cloud.bigquery import LoadJob, LoadJobConfig, QueryJob, SourceFormat, CopyJobConfig, CopyJob
+from google.cloud.bigquery import LoadJob, LoadJobConfig, QueryJob, SourceFormat, CopyJobConfig, CopyJob, dataset
 from google.cloud.bigquery.job import QueryJobConfig
 from google.cloud.exceptions import Conflict, NotFound
-from google.cloud.storage import Blob
+from google.cloud.storage import Blob, bucket
 from googleapiclient import discovery as gcp_api
 from observatory.api.client.model.big_query_bytes_processed import (
     BigQueryBytesProcessed,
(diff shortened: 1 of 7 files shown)


alexmassen-hane:
Added new unit tests for new functions. All buckets and datasets that are created for unit tests should have a prefix with an abbreviation of at least the class name, and the organisation name if available.

observatory-platform/observatory/platform/utils/gc_utils.py
```diff
@@ -1232,3 +1234,136 @@ def delete_bucket_dir(*, bucket_name: str, prefix: str):
...
+    # Client to read Google Cloud Storage
+    storage_client = storage.Client()
+    buckets = storage_client.list_buckets()
+
+    # Make list of all names of the storage buckets
+    bucket_list = []
+    for bucket in buckets:
+        bucket_list.append(bucket.name)
+
+    return bucket_list
```
jdddog, line 1254: This function and `get_age_of_bucket_in_days` could be consolidated by directly returning the list of `Bucket` instances, since `list_buckets` already returns Bucket instances, and calling `get_age_of_bucket_in_days` effectively means that they are fetched twice.

Assuming that `list_buckets` is a generator, the code from 1247 to the end of the function could be replace by `list(storage_client.list_buckets())`. You can then perform the time difference calculation, calling `bucket.time_created` from within `delete_old_buckets_with_prefix`.

It would also be good to specify the return type of the function:
```
def list_all_buckets() -> List[Bucket]:
```

Bucket is the type `google.cloud.storage.bucket.Bucket`.

observatory-platform/observatory/platform/utils/gc_utils.py
```diff
@@ -1232,3 +1234,136 @@ def delete_bucket_dir(*, bucket_name: str, prefix: str):
...
+    """
+
+    client = bigquery.Client()
+    datasets = list(client.list_datasets())
+
+    dataset_list = []
+    for dataset in datasets:
+        dataset_list.append(dataset.dataset_id)
+
+    return dataset_list
```
jdddog, line 1270: I would also try to consolidate `list_all_datasets` and `get_age_of_dataset_in_days`, i.e. make `list_all_datasets` return a list of `Dataset` objects, and then put the time difference calculation in `delete_old_datasets_with_prefix` (which calls calls `dataset.created`).

observatory-platform/observatory/platform/utils/gc_utils.py
```diff
@@ -1232,3 +1234,136 @@ def delete_bucket_dir(*, bucket_name: str, prefix: str):
...
+    client = storage.Client(project=project_id)
+
+    # List all buckets in the project.
+    bucket_list = list_all_buckets()
+
+    buckets_deleted = []
+    for bucket_id in bucket_list:
+
+        # Check if prefix is in the bucket name.
+        if (prefix in bucket_id) and (prefix != ""):
```
jdddog, line 1320: It would be safer to call `bucket_id.startswith(prefix)` so that only buckets that start with the prefix are deleted, not buckets with the prefix anywhere in their name.

observatory-platform/observatory/platform/utils/gc_utils.py
```diff
@@ -1232,3 +1234,136 @@ def delete_bucket_dir(*, bucket_name: str, prefix: str):
...
+    client = bigquery.Client(project=project_id)
+
+    # List all datsets in the project.
+    dataset_list = list_all_datasets()
+
+    datasets_deleted = []
+    for dataset_id in dataset_list:
+
+        # Check if prefix is in the dataset name.
+        if (prefix in dataset_id) and (prefix != ""):
```
jdddog, line 1357: It would be safer to call `dataset_id.startswith(prefix)` so that only datasets that start with the prefix are deleted, not datasets with a prefix anywhere in the string.

observatory-platform/observatory/platform/utils/gc_utils.py
```diff
@@ -1232,3 +1234,136 @@ def delete_bucket_dir(*, bucket_name: str, prefix: str):
...
+
+    client = bigquery.Client(project=project_id)
+    dataset = client.get_dataset(dataset_id)
+
+    # Timezone of the dataset is in UTC. Have to change local timezone to UTC to get correct age.
+    days_old = (datetime.datetime.now(datetime.timezone.utc) - dataset.created).days
+
+    return days_old
+
+def delete_old_buckets_with_prefix(project_id: str, prefix: str, age_to_delete: int = 7):
```
jdddog, line 1303: It would be good to use hours rather than days, as leaving it for days could build up a large amount of buckets.

observatory-platform/observatory/platform/utils/gc_utils.py
```diff
@@ -1232,3 +1234,136 @@ def delete_bucket_dir(*, bucket_name: str, prefix: str):
...
+                buckets_deleted.append(bucket_id)
+
+    if len(buckets_deleted) < 1:
+        logging.info(f"No buckets older than {age_to_delete} days to delete.")
+    else:
+        logging.info(f"Deleted the following buckets older than {age_to_delete} days: {buckets_deleted}")
+
+    return buckets_deleted
+
+def delete_old_datasets_with_prefix(project_id: str, prefix: str, age_to_delete: int = 7):
```
jdddog, line 1340: It would be good to use hours rather than days, as leaving it for days could build up a large amount of datasets.

observatory-platform/observatory/platform/utils/test_utils.py
```diff
@@ -166,13 +171,84 @@ def save_empty_file(path: str, file_name: str) -> str:
...
+        prefix = (
+        ("".join(re.findall("[A-Z]+", test_name))).lower()
+        )
+
+    logging.info(f"Prefix for buckets and datsets: {prefix}")
+
+    return prefix
+
+
 class ObservatoryEnvironment:
```
jdddog, line 244: - It would be good if you didn't have to specify the prefix when creating `ObservatoryEnvironment`. To make it simple we could have a standard global prefix, defined as a keyword argument in the constructor, e.g. "obs_tests". It is OK if one test deletes another test's datasets or buckets, as long as enough time has passed to know that they are not still being used.

- We should probably also modify `add_bucket` to have an optional prefix parameter like `add_dataset` so that the test writer can make buckets more easily identified, we just need to be careful / have checks because of the bucket character limit.

- We could call `delete_old_test_datasets` and `delete_old_test_buckets` from within the `create` function, in the finally statement, after line 645, then the datasets and buckets will always be removed and we don't need to add extra code to all existing unit tests (probably just the tests added in test_gc_utils) :

```
    if self.create_gcp_env:
        # Remove Google Cloud Storage buckets
        for bucket_id in self.buckets:
            self._delete_bucket(bucket_id)
        # Remove BigQuery datasets
        for dataset_id in self.datasets:
            self._delete_dataset(dataset_id)

       # Call delete_old_test_datasets and delete_old_test_buckets here
```

* Lastly, the above change means that we could put the `age_to_delete` variable as a keyword argument in the constructor of `ObservatoryEnvironment`. Perhaps set as 1 or 2 hours?

jdddog (changes requested):
Hey Alex, it is looking pretty good. At a high level it, would be good to consolidate the functions that call Google Cloud, and move some of the parameters and function calls into the `ObservatoryEnvironment` class so that the unit test writer doesn't need to add extra functionality themselves. Also just make sure that you run the Black formatter on all of the files that have changed.

alexmassen-hane:
Changes:
- Added additional prefix option to add_bucket, like in add_dataset, and will raise an exception if it is more than 63 characters for a bucket name (Google's limit).
- Removed duplicate functions and tests for `delete_old_test_buckets` and `delete_old_test_datasets` in the ObservatoryEnvironment class.
- Observatoryenvironment class only calls `delete_old_test_buckets_with_prefix` and `delete_old_test_datasets_with_prefix` in the `create` function.
- Made global prefix in the ObservatoryEnvironment class constructor, "obsenv_tests", as well with the "age_to_delete" variable. 
- Changed age check to hours instead of days, and set default for "age_to_delete" to 2 hours since unit tests typically take about 1-1.5 hours on github.
- `list_all_buckets` and `list_all_datasets` now returns list of objects, also specifies output of the functions (list of objects).
- Removed `get_age_of_datasets` and `get_age_of_bucket` functions and moved age check to inside `delete_old_test_buckets_with_prefix` and `delete_old_test_datasets_with_prefix`.

alexmassen-hane:
Changes:
- list_all datasets and bucket functions now only list those with a prefix to get around a logic error when there are parallel unit tests running. This is to prevent tests from trying to retrieve buckets/datasets objects that have just been deleted by congruent tests from the Google API. This is to avoid them failing when being tested on Github.
- New functions cause tests to run for longer so they may potentially delete buckets/datasets that are still in use but over age_to_delete time value. Have increased the default age_to_delete value from 2 to 12 hours.

observatory-platform/observatory/platform/utils/test_utils.py
```diff
@@ -243,18 +252,31 @@ def assert_gcp_dependencies(self):
 
         assert self.create_gcp_env, "Please specify the Google Cloud project_id and data_location"
 
-    def add_bucket(self) -> str:
+    def add_bucket(self, prefix: str = "") -> str:
```
jdddog, line 255:
```suggestion
    def add_bucket(self, prefix: Optional[str] = None) -> str:
```

observatory-platform/observatory/platform/utils/test_utils.py
```diff
@@ -243,18 +252,31 @@ def assert_gcp_dependencies(self):
...
-        bucket_name = random_id()
-        self.buckets.append(bucket_name)
+        if (self.prefix != "") and (prefix != ""):
+            bucket_name = f"{self.prefix}_{prefix}_{random_id()}"
+        elif self.prefix != "" and (prefix == ""):
+            bucket_name = f"{self.prefix}_{random_id()}"
+        elif (self.prefix == "") and (prefix != ""):
+            bucket_name = f"{prefix}_{random_id()}"
+        else:
+            bucket_name = random_id()
```
jdddog, line 273: This is an alternative way to make the prefix which might be easier to follow (an empty string evaluates to False as does None):

```python
parts = []
if self.prefix:
  parts.append(self.prefix)
if prefix:
  parts.append(prefix)
parts.append(random_id())
bucket_name = "_".join(parts)
```

observatory-platform/observatory/platform/utils/test_utils.py
```diff
@@ -302,12 +326,16 @@ def add_dataset(self, prefix: str = "") -> str:
...
         self.assert_gcp_dependencies()
-        if prefix != "":
+        if (self.prefix != "") and (prefix != ""):
+            dataset_id = f"{self.prefix}_{prefix}_{random_id()}"
+        elif self.prefix != "" and (prefix == ""):
+            dataset_id = f"{self.prefix}_{random_id()}"
+        elif (self.prefix == "") and (prefix != ""):
             dataset_id = f"{prefix}_{random_id()}"
         else:
             dataset_id = random_id()
```
jdddog, line 341: This is an alternative way to make the prefix which might be easier to follow (an empty string evaluates to False as does None):

```python
parts = []
if self.prefix:
  parts.append(self.prefix)
if prefix:
  parts.append(prefix)
parts.append(random_id())
bucket_name = "_".join(parts)
```

You would also want to change the prefix parameter to `prefix: Optional[str] = None`

observatory-platform/observatory/platform/utils/test_utils.py
```diff
@@ -179,6 +182,8 @@ def __init__(
         enable_elastic: bool = False,
         elastic_port: int = 9200,
         kibana_port: int = 5601,
+        prefix: str = "obsenv_tests",
```
jdddog, line 185:
```suggestion
        prefix: Optional[str] = "obsenv_tests",
```

jdddog (changes requested):
Hey Alex, it is looking good, just a few small comments.

jdddog (approved):
Thanks Alex, looks good!

PR #579, opened 2022-09-01
precommit rebase fix

keegansmith21:
The pre-commit hook would run into issues during an interactive rebase. The new command fixes the issue but noticing when a rebase is taking place and skipping the commit hook


jdddog (approved):
Thanks Keegan, it is good to have this fixed!

PR #584, opened 2022-09-27
INF-461: Exception handling for deleting datasets and buckets. 

alexmassen-hane:
Needing to add exception handling to the new functions that delete old test datasets and buckets as tests that are run in parallel try to delete the same objects at once and will fail if they are already deleted.

Added try and catch in the new functions where necessary to avoid unit tests failing.


jdddog (approved):
Thanks for the fix Alex!

jdddog (changes requested):
sdsd

PR #585, opened 2022-09-30
INF-558: Added exception handling for deleting datasets and buckets produced by unit tests

alexmassen-hane:
In-between the time that unit tests list and then grab the objects for datasets and buckets from the Google API, they could have already been deleted by another unit test. This results in a "Not Found" error and the unit tests fail. 

Have added appropriate error handling for if they don't exist before grabbing the objects from the API and before they are deleted.


jdddog (approved):
Thanks Alex, it looks good!

PR #587, opened 2022-10-09
Increase BQ bytes budget for new DOI workflow

jdddog:
The updated DOI workflow uses more bytes in a single query and more bytes for the entire day, so I have increased the limits for both.


keegansmith21 (approved):
Looks good! Thanks Jamie!

PR #588, opened 2022-10-19
Log BulkIndexErrors

jdddog:
Detailed errors are not currently printed when exceptions occur indexing documents in Elasticsearch, which makes it hard to figure out what the problem is. Updated the code to log each error in a BulkIndexError.


keegansmith21 (approved):
This will be helpful, we get a lot of import errors! Thanks Jamie

PR #589, opened 2022-11-02
BAD-308 schema and table description updates

keegansmith21:
This update was initially scoped to fix an issue where a table's descriptions are wiped if the table is updated. 
In looking through the foundational utility functions, the following things have been changed:

# find_schema()
## The function has been altered in the following ways:
- Now returns dateless schema files where a date is not supplied. The date is now an optional argument
- Removed the versioning utility of the function as it is not used in any case other than in tests (which have been updated accordingly)

## This results in the following changes to underlying code:
- The schema finding function has been abstracted out of the underlying table creation, making it more obvious where the schema is coming from during table creation. 
- Schema files should now be explicitly supplied to the table creation functions. This gives the user more control over what schema is being supplied at the telescope/workflow level. This is significantly more important now as we are forcing more tables to comply with the supplied schema. Some cases disallow this use however, such as the [stream release telescope's partition load ](https://github.com/The-Academic-Observatory/observatory-platform/blob/dfdfcd48595d0f0595cd9f26fc60dd7c38b36ffb/observatory-platform/observatory/platform/workflows/stream_telescope.py#L262). This could be changed, but would require reworking some parts of the telescope.

# create_bigquery_table_from_query()
## The function has been altered in the following ways:
- If a schema file is supplied, any existing table with the supplied name will be deleted before being recreated

## This results in the following behaviour changes when calling the function
- This ensures that the table descriptions are updated when a table exists already and is being updated with a schema. Since bigquery will wipe existing descriptions when updating a table (for some reason).

# Table loading functions
### These functions have changed in the following ways:
1. Removed prepare_bq_load and prepare_bq_load_v2
2. Merged the functionality of the two bq_load_shard functions into bq_load_shard()

### This was done for the following reasons:
1. prepare_bq_load was used to wrap the find_schema function and create_bigquery_dataset. Since find_schema has been abstracted out, this meant it was only wrapping a single function, making it redundant.
2. The two bq_load_shard functions did very similar things but gathered the necessary information differently. I have forced the information to be supplied explicitly in all cases. This removes duplication of code and also makes it more clear what tables and datasets are being made at the telescope/workflow level

### These functions have changed in the following ways:
- Each function now expects a supplied schema, project_id, bucket and data_location

### This was done for the following reasons:
- Schema files should now be defined more explicitly wherever possible for reasons given above
- project_id, bucket and data_location were normally inferred from prepare_bq_load(). They need to be supplied in a different manner now. I think that this way is better as this information is always available to the telescope/workflow anyway and should make it more clear where the table is going to be created.

# Relates to:
- The-Academic-Observatory/oaebu-workflows#116
- The-Academic-Observatory/academic-observatory-workflows#145


jdddog (changes requested):
Hey Keegan, the changes look good, just a couple of points for feedback:

* I think we should change the logic so that if you don't supply a release_date, then it just finds the schema with the exact table name. Then if you supply the release date then it finds the latest version of the schema. That would make the logic easier to understand.
* How are schemas with versions handled? Some workflows in the academic observatory have a schema_version parameter. I don't know if they are used, maybe we could just remove them.

tests/observatory/platform/utils/test_config_utils.py
```diff
@@ -71,83 +71,39 @@ def test_terraform_credentials_path(self):
...
         self.assertIsNotNone(result)
         self.assertTrue(result.endswith(expected_schema))
 
-        # 2020-04-28
-        expected_schema = "grid_2016-04-28.json"
-        release_date = pendulum.datetime(year=2016, month=4, day=28)
-        result = find_schema(schemas_path, table_name, release_date)
+        # Prefix test
+        expected_schema = "prefix_table.json"
+        result = find_schema(schemas_path, "table", prefix="prefix_")
```
jdddog, line 88: It would be good to add a test back that finds a table with a prefix, where there are two available dates, which is what the previous tests did.

The tests should check all of the combinations of: prefix, table_name and release_date.

jdddog (approved):
Hey Keegan, these changes look good to me, nice work!

PR #592, opened 2022-11-28
Update github unit tests in github workflow/actions

keegansmith21:
Removed the "on pull request" unit test as it's redundant


jdddog (changes requested):
Hey Keegan, thanks for the Github Actions PRs. 

We are using two spaces for YAML indentation, would you be able to reformat it with 2 spaces? It would be good to keep the same number of spaces as this makes it easier to go back and compare changes.

It should really just be lines 3-8 that change to: `on: push`

PR #594, opened 2022-12-05
API core version

jdddog:
Just checking if the tests pass first.


keegansmith21 (approved):
Thanks for fixing this Jamie!

PR #596, opened 2023-01-16
View creation function update

keegansmith21:
This is a PR to alter the functionality of the the view creation function in gc_utils.py

Previously, if you attempted to use create_bigquery_view using a view that already exists, nothing would happen. This is becuase client.create_table() cannot be used to update a table, and since the exist_ok tag was used, no error is thrown. 

I believe that this is not the best behaviour for the function, so I have modified it with an optional input that, when true, will attempt to update the view (specifically, the view's query) if it already exists. 

I have updated the tests to include this change


alexmassen-hane (approved):
Looks good Keegan!

PR #597, opened 2023-01-24
INF-588: HTTP Request Updates

keegansmith21:
Updates to URL Utilities

- Removed functions that are not in use anywhere
- Created a new function for retrying GET requests using tenacity

I know that we have the retry_session() function, but there are some things that it (and urllib3) can't do. With the tenacity function, we can:
- specify any 400/500 response as an error
- log each repeat request as they are placed

This does not make the retry_session() function obsolete, as it is able to force a small selection of response codes to retry. It can also do more than GET requests.

### DIFF
--- a/observatory-platform/observatory/platform/utils/url_utils.py
+++ b/observatory-platform/observatory/platform/utils/url_utils.py
@@ -14,69 +14,70 @@
 
 # Author: James Diprose
 
-import hashlib
 import json
 import os
 import time
 import urllib.error
 import urllib.request
 from typing import Any, Dict, List, Tuple, Union
-from urllib.parse import ParseResult, urljoin, urlparse
+import logging
 
 import requests
-import tldextract
 import xmltodict
 from importlib_metadata import metadata
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
+from tenacity import Retrying, stop_after_attempt, before_sleep_log, wait_exponential_jitter
+from tenacity.wait import wait_base
 
 
-def get_url_domain_suffix(url: str) -> str:
-    """Extract a URL composed of the domain name and the suffix of the URL. For example, library.curtin.edu would
-    become curtin.edu
+def retry_get_url(
+    url: str,
+    num_retries=3,
+    wait: wait_base = wait_exponential_jitter(multipliter=5, max=300),
+    squelch_url: bool = False,
+    **kwargs,
+):
+    """Attempts an HTTP GET request inside a retry loop.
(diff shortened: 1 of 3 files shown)


observatory-platform/observatory/platform/utils/url_utils.py
```diff
@@ -14,69 +14,53 @@
...
+        with attempt:
+            response = requests.get(url, headers=headers)
+            if response.status_code != 200:
+                logging.info(
+                    f"Downloading events from url: {log_url} returned response code {response.status_code} "
+                    f"attempt: {attempt.retry_state.attempt_number} - idle for: {attempt.retry_state.idle_for}"
+                )
+                raise ConnectionError(
+                    f"Got response code {response.status_code}, response: {response.text}, reason: {response.reason}"
+                )
```
jdddog, line 58: It would be good to make the logged message and the ConnectionError have the same message, which would involve adding the response.reason to the error message.

Also the logging error should be more generic - it still has text from Crossref Events, e.g. "Error getting url: " rather than "Downloading events from url"

What happens if requests throws an exception? (see first answer)

https://stackoverflow.com/questions/16511337/correct-way-to-try-except-using-python-requests-module

jdddog (changes requested):
Hey Keegan, looks good, just a few comments.

observatory-platform/observatory/platform/utils/url_utils.py
keegansmith21: Hi Jamie. I've made some changes to the function that I think make it much cleaner.

requests.get() generally doesn't return an error, so I've added raise_for_status() which will raise an error for any 400/500 return code. This will then be caught, relevant information will be extracted and it will then be reraised. This will be caught by tenacity, which would normally not log an error when retrying, but before_sleep_log() will write the error message to the logger. I've also added a timeout parameter, which will immediately raise an error should the request take too long, which is why the error message changes slightly when a timeout occurs.

I also removed the response.text log as it can be quite lengthy.

The resulting retry logs look like this:
```INFO:__main__:Retrying __main__.retry_get_url in 1.0 seconds as it raised RequestException: Error getting url: http://example.com/frgfds | Got response code: 500 | Reason: Internal Server Error | Attempt: 2 | Idle for: 1.0.```

For timeout:
```INFO:__main__:Retrying __main__.retry_get_url in 1.0 seconds as it raised ReadTimeout: Error getting url: http://example.com/frgfds | Attempt: 1 | Idle for: 0.0.```

jdddog (approved):
Thanks Keegan, the changes look good.

PR #598, opened 2023-02-14
INF-597: Terraform config files not updating and Flower docker container not building successfully. 

alexmassen-hane:
1. When using the `observatory terraform build-image config.yaml` command, it was not copying/overwriting terraform config files to the observatory home directory and thus when deploying a new version of the platform to the cloud, it was using the old terraform config. I have now forced the platform to delete the old terraform config files before copying the new ones across from the codebase.

3. After updating the terraform files a strange permissions issue with docker arose, causing the flower container to fail when building the main VM. Have added the appropriate permissions to the /home/packer folder for the airflow user so the Flower container can build successfully. 

Waiting on results from the unit tests.

### DIFF
--- a/observatory-platform/observatory/platform/terraform/terraform_builder.py
+++ b/observatory-platform/observatory/platform/terraform/terraform_builder.py
@@ -31,8 +31,8 @@ from observatory.platform.utils.jinja2_utils import render_template
 from observatory.platform.utils.proc_utils import stream_process
 
 
-def copy_dir(source_path: str, destination_path: str, ignore, update: int = 0):
-    distutils.dir_util.copy_tree(source_path, destination_path, update=update)
+def copy_dir(source_path: str, destination_path: str):
+    distutils.dir_util.copy_tree(source_path, destination_path)
 
 
 class TerraformBuilder:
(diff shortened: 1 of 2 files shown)


observatory-platform/observatory/platform/terraform/terraform_builder.py
```diff
@@ -31,8 +31,8 @@
 from observatory.platform.utils.proc_utils import stream_process
 
 
-def copy_dir(source_path: str, destination_path: str, ignore):
-    distutils.dir_util.copy_tree(source_path, destination_path)
+def copy_dir(source_path: str, destination_path: str, ignore, update: int = 0):
+    distutils.dir_util.copy_tree(source_path, destination_path, update=update)
```
jdddog, line 35: It looks like the `update` parameter is already set to zero by default in `distutils.dir_util.copy_tree`, so we shouldn't need to set it.

I seem to have messed up some of this code, because the ignore parameter doesn't get passed to anything. Maybe we were originally using `shutil.copytree` as that has an ignore parameter? We should probably either remove the ignore parameter or figure out if we could use shutil.copytree which can ignore files with patterns.

jdddog (commented):
Thanks for the fix Alex, that's great. Just a small comment about one of the functions.

jdddog (approved):
Thanks Alex, looks good.

PR #599, opened 2023-02-15
Remove project and workflow generation

jdddog:
Remove project and workflow generation to reduce maintenance burden, especially since we are changing how some of the workflows will be created.


keegansmith21 (approved):
Thanks for these changes Jamie! :smile:

PR #600, opened 2023-02-15
Feature/remove query api

jdddog:
Removing reports / query endpoint from API as it is broken and not being used. Regenerated API client.

We should make a new release for the API image and update the URL on the next deployment.


keegansmith21 (approved):
These changes look good Jamie. Thanks for trimming down on maintenance work!

PR #602, opened 2023-02-16
Upgrade to Docker Compose V2

jdddog:
Upgrade Docker Compose to V2 as V1 won’t be supported from the end of June 2023 and will be removed from all Docker Desktop versions.

I've also updated a number of related things:
* Updated docs with Docker Compose V2 changes and Ubuntu install script.
* For VMs:
  * Updated VM image to Ubuntu 22.04 (Ubuntu 18.04 is end of life).
  * Updated Berglas to v1.0.1.
* Terraform:
  * Tidied up the firewall rules, which had to be updated after hashicorp/google updated.
  * Add versions restraints to the terraform/versions.tf so that the Terraform providers don't update to a new version without us doing it, which causes problems.

I've tested deploying the system to the cloud.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -5,6 +5,7 @@ Observatory Platform supports Python 3.8, Ubuntu Linux 20.04 and MacOS 10.14, on
 * Python 3.8
 * Pip
 * Docker
+* Docker Compose V2
 * virtualenv
 * curl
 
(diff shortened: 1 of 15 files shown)


keegansmith21 (approved):
Thanks for these changes Jamie!

PR #603, opened 2023-02-16
Remove Elastic and Kibana from local platform

jdddog:
Remove Elastic and Kibana from local platform as they are not used. Also removed some documentation about the Terraform deployment related to the query API that we removed.


keegansmith21 (approved):
Thanks for these changes Jamie. It will be nice to not have these containers eating my RAM!

PR #604, opened 2023-02-20
Fix: Only delete non hidden files in terraform directory. 

alexmassen-hane:
@jdddog Requested this fix for the merged pull request INF-597: Terraform config files not updating and Flower docker container not building successfully. 

The .terraform folder in the /build/terraform/terraform directory is needed to be kept and not deleted each time the terraform part of the platform initialises. 

I have changed it so instead of deleting the path /build/terraform/terraform, it will now only delete all of the non hidden files in that directory, keeping the necessary .terraform folder and .terraform.lock.hcl file.


jdddog (changes requested):
Thanks Alex, it looks good. The only thing to change is to use `os.path.join` rather than constructing the paths like this `f"{self.terraform_build_path}/{file}"` as it will not work across different operating systems.

jdddog (approved):
Thanks Alex, looks good!

PR #607, opened 2023-02-24
Move helper functions from Thoth Telescope to common Observatory Platform utilities

alexmassen-hane:
Move helper functions from the Thoth Telescope to the main Observatory Platform Utilities.


jdddog (approved):
Thanks Alex, it looks good.

PR #608, opened 2023-02-28
Change table_id to table_name

jdddog:
Change table_id to table_name to make it clear that it isn't the full table id that needs to be supplied to the function.


keegansmith21 (approved):
Thanks for these changes Jamie!

PR #610, opened 2023-03-08
INF-465: Cloud Endpoints Portal Deprecation - move observatory-api container

alexmassen-hane:
On March 21st, the Developer portal for Google APIs is being deprecated and we will not have access to a production API gui where we can easily check or modify Observatory API data. 

This PR is to move the deployment of the observatory-api container from Google's Cloud Run service to a local docker container that runs on the main and worker VM of the deployed Observatory Platform. In doing so, we will be able to access the Swagger UI by ssh'ing into the main VM and going to http://localhost:5002/ui . The local observatory-api container will still have access to the Cloud DB instance, where all of the release and workflow data is stored.  

The API gateway and the serverless VPC service have also been removed as they were only needed for the Cloud Run instance to access data from the Cloud DB instance that holds the observatory API data. 

TODO: 
- decide on a convention of how to access the local observatory-api container, docker-apiserver:5002 or localhost:5002
- decide on if we still want to do a 'build-image' of the platform and have it stored in Google's container registry
- decide if we want to remove the "api" configuration from the config.yaml format.


observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
```diff
@@ -16,6 +16,7 @@
 version: '3.8'
 
 x-environment: &environment
+
```
jdddog, line 19:
```suggestion
```

jdddog (commented):
Thanks Alex, your PR looks good, I've just got a few small suggestions.

These files can be removed:
* .github/workflows/build-observatory-api.yml
* observatory-api/Dockerfile
* observatory-api/requirements.cloudrun.txt

Then in `terraform_builder.py`, the function `make_open_api_template` and any calls to it can be removed as well as the import for OpenApiRenderer. This isn't needed anymore since the openapi.yaml file is rendered when the API server is started.

PR #612, opened 2023-03-09
Fix/dependencies

jdddog:
Fixed an issue with platform starting and dependencies getting installed. Addressed by fixing an issue in requirements.txt, and using stock Airflow images for flower, webserver and adding the airflow_init version container. Tidied up the Docker Compose file a little bit and migrated from Google Compute Monitoring agent to Google Compute Ops Agent.


alexmassen-hane (approved):
Looks good, thanks Jamie.

PR #615, opened 2023-03-24
Fix/Change apiserver container docker network

alexmassen-hane:
Connecting to the apiserver container by hostname was not working because the docker network details were incorrect when deployed to the cloud. This change is required so that container can report back as 'healthy' and so Airflow can use it for all of the workflows.


jdddog (approved):
Thanks Alex!

PR #616, opened 2023-03-29
INF-609: Add limits to Bigquery for per user per day and project per day. 

alexmassen-hane:
Adding safety limits for Bigquery usage for per user per day and per day for the entire Google project that the observatory platform is deployed to. 

Current limits are set to 10 TiB ( 10,485,760 Megabytes ) for the per user per day and 15 Tib ( 15,728,640 Megabytes ) per day for the whole project. 

We can easily increase these limits if required.


keegansmith21 (approved):
Thanks Alex, I like this!

jdddog (approved):
Thanks Alex, this is a really useful feature!

PR #617, opened 2023-03-30
Fix api and workers network settings

jdddog:
Fixes the issues with Airflow webserver not being able to access logs and API server not being accessible.

Made following changes:
* Reverted workers and apiserver to use host mode. API and worker containers are now accessed using the host name of the VM.
* In Terraform:
  * Open ports 5002 (API) and 8793 (Airflow worker) within VPC network for both VMs. 
  * Only run apiserver on airflow-main-vm.
  * Use hostname of airflow-main-vm for redis_hostname instead of IP address.
* observatory_api connection should now be: http://:@airflow-main-vm:5002


alexmassen-hane (dismissed):
Thanks Jamie. It's a shame we couldn't keep the apiserver container on each main and worker VM. I'll note down the structure of the containers in the technical overview document.

PR #623, opened 2023-06-22
Changed default write disposition

keegansmith21:
Default table load probably shouldn't be destructive (as I've discovered the hard way).


jdddog (approved):
Thanks Keegan, that looks like a safer option. I've updated a couple of workflows in Academic Observatory to work in line with this: https://github.com/The-Academic-Observatory/academic-observatory-workflows/pull/170

PR #624, opened 2023-06-28
Added function to list blobs in gcs bucket

keegansmith21:
I've found use for a function like this in the academic observatory. I thought it would be better to put it here.


jdddog (approved):
Thanks Keegan, looks good!

PR #625, opened 2023-07-06
Feature/Match on multiple keys for upserts and deletes in Bigquery. 

alexmassen-hane:
Add functionality to use either a singular key or a list of keys to match on for upserts and deletes in Biguqery. I have added this as a union of a string and list so we do not have to modify every workflow that uses these two functions. 

There's probably a much cleaner way of changing the jinja sql scripts for this. 
Please review and suggest other ways if possible.

### DIFF
--- a/observatory-platform/observatory/platform/sql/delete_records.sql.jinja2
+++ b/observatory-platform/observatory/platform/sql/delete_records.sql.jinja2
@@ -17,13 +17,9 @@
 
 MERGE `{{ main_table_id }}` main_table
 USING `{{ delete_table_id }}` delete_table
-{% if main_table_keys|length == 1 %}
-ON CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_keys[0] }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_keys[0] }})
-{% else %}
-ON ( CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_keys[0] }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_keys[0] }}) AND
-{% for (main_table_key, delete_table_key) in zip(main_table_keys[1:-1], delete_table_keys[1:-1]) %}
-   CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_key }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_key }}) AND
-{% endfor %}
-   CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_keys[-1] }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_keys[-1] }}) )
-{% endif %}
+ON (
+{%- for (main_table_key, delete_table_key) in zip(main_table_keys, delete_table_keys) %}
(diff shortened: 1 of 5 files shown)


keegansmith21 (dismissed):
Looks good! Thanks Alex! :+1:

observatory-platform/observatory/platform/sql/upsert_records.sql.jinja2
```diff
@@ -14,7 +14,15 @@
 
 MERGE `{{ main_table_id }}` main_table
 USING `{{ upsert_table_id }}` upsert_table
-ON main_table.{{ primary_key }} = upsert_table.{{ primary_key }}
+{% if keys|length == 1 %}
```
jdddog, line 17: The jinja template could be simplified a little bit, I've also remove the newlines in the output (which were already there) with the minus signs:

```jinja2
MERGE `{{ main_table_id }}` main_table
USING `{{ upsert_table_id }}` upsert_table
ON (
  {%- for key in keys %}
  main_table.{{ key }}=upsert_table.{{ key }}{% if not loop.last %} AND{% endif %}
  {%- endfor %}
)
WHEN MATCHED THEN
  UPDATE SET
    {%- for col in columns %}
    main_table.{{ col }}=upsert_table.{{ col }}{% if not loop.last %},{% endif %}
    {%- endfor %}
WHEN NOT MATCHED THEN
  INSERT ROW
```

Then the output would be:
```sql
MERGE `my-project.bq_tests_c9b656431a145bdb8b9c22130bbf81f1.presidents` main_table
USING `my-project.bq_tests_c9b656431a145bdb8b9c22130bbf81f1.presidents_upserts` upsert_table
ON (
  main_table.name=upsert_table.name AND
  main_table.death_age=upsert_table.death_age
)
WHEN MATCHED THEN
  UPDATE SET
    main_table.name=upsert_table.name,
    main_table.death_age=upsert_table.death_age,
    main_table.date=upsert_table.date
WHEN NOT MATCHED THEN
  INSERT ROW
```

observatory-platform/observatory/platform/sql/delete_records.sql.jinja2
```diff
@@ -17,5 +17,13 @@
...
-ON CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_primary_key }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_primary_key }})
+{% if main_table_keys|length == 1 %}
+ON CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_keys[0] }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_keys[0] }})
+{% else %}
+ON ( CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_keys[0] }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_keys[0] }}) AND
+{% for (main_table_key, delete_table_key) in zip(main_table_keys[1:-1], delete_table_keys[1:-1]) %}
+   CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_key }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_key }}) AND
+{% endfor %}
+   CONCAT('{{ main_table_primary_key_prefix }}', main_table.{{ main_table_keys[-1] }}) = CONCAT('{{ delete_table_primary_key_prefix }}', delete_table.{{ delete_table_keys[-1] }}) )
+{% endif %}
```
jdddog, line 28: To simplify the jinja template, try to use the approach in the suggestion for the upsert table here.

jdddog (changes requested):
Thanks Alex, it looks good, I just have a suggestion for simplifying the upsert and delete jinja templates.

jdddog (approved):
Thanks Alex, the changes look good. I think unrelated tests are failing because they are flaky, lets see if they pass this time...

PR #626, opened 2023-07-07
Added multi-uri table load

keegansmith21:
I noticed that the table load functionality (on google's end) allows for glob pattern matched URIs. However the list_blobs() functionality (again on google's end) does not! This poses an issue when you want to check that the table load URIs match what you expect when you use list_blobs(). So to get around this, we can just manually supply a list of URIs so that we're certain of the blobs being loaded.


jdddog (approved):
Thanks Keegan, it will be very useful to be able to manually pass a list of URIs for loading data into tables.

PR #627, opened 2023-07-30
Upgrade to Airflow 2.6.3

jdddog:
This PR upgrades our system to use Airflow 2.6.3.

An important point to be aware of, is that when you install our packages, you need to add the `--constraint` argument:
```
pip install -e observatory-api[tests] --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.6.3/constraints-no-providers-3.8.txt
pip install -e observatory-platform[tests] --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.6.3/constraints-no-providers-3.8.txt
```

This applies for the workflow packages as well. It makes sure that the correct versions of the packages that Airflow depends on that our projects also depend on, are installed. I've added this to our install script as well.

Other points:
* Remove the Elastic Import Workflow and Elastic / Kibana testing environment as these are not being used anymore and removing them will reduce maintenance. 
* The Airflow `schedule_interval` parameter is now called `schedule`.
* Some of the tests in the VM destroy workflow are failing because they are publishing XComs outside of a DAG run and constraints introduced in a newer Airflow version prevent this. We may remove this workflow shortly if we move to an Airflow managed service, so rather than spending time fixing these tests, I have annotated them with `@unittest.skip`.


keegansmith21 (approved):
Thanks Jamie. Changes look good!

alexmassen-hane (approved):
Looks good!

PR #629, opened 2023-08-08
Fix Google Cloud Storage based logging

jdddog:
Provides packages required to enable Google Cloud based Airflow logging and stops certain files from being uploaded to the VM image.


keegansmith21 (approved):
This looks good. Thanks Jamie.

PR #630, opened 2023-08-14
Fix/Logs disappear when tasks are in "up for retry" state

alexmassen-hane:
With the upgrade to Airflow 2.6, logs will disappear in the web UI when a task's status is in the "up for retry" stat,e however when they are in the "failed" or "success" state the logs are available. This is due to the Airflow web UI only finding the logs on the local and remote celery workers through the docker network and there is no open connection for the webserver container to find the logs when in the "up for retry" state. 

This PR forces the common data and logs volumes to be mapped to the webserver container. This allows the web UI to find the logs locally instead of sourcing them from the worker containers in the docker network.

I have also added user-friendly container and host names to the containers for easy debugging down track.

### DIFF
--- a/observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
+++ b/observatory-platform/observatory/platform/docker/docker-compose.observatory.yml.jinja2
@@ -109,6 +109,8 @@ x-build: &build
 
 services:
   redis:
+    container_name: redis
+    hostname: redis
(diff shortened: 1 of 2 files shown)


keegansmith21 (approved):
Thanks Alex, looks good to me!

jdddog:
Thanks for fixing this and for testing it on the cloud deployed version Alex!

PR #631, opened 2023-08-16
Make dag_run_id nullable in openapi.yaml

jdddog:
A fix for the error:

```python
observatory.api.client.exceptions.ApiTypeError: Invalid type for variable 'dag_run_id'. Required value type is str and passed type was NoneType at ['received_data'][0]['dag_run_id']
```


keegansmith21 (approved):
Thanks for getting on to this Jamie!

PR #632, opened 2023-08-16
hmac key

keegansmith21:
I have recently required the use of an HMAC key. I thought that I should create one and add it as an airflow connection, but discovered that you can just make one on the fly using your service account credentials. So I made this - which also deletes the key when the context manager closes.


jdddog (approved):
Thanks Keegan, it looks good.

PR #633, opened 2023-08-17
Added glob match functionality to list_blobs

keegansmith21:
Added the ability to match on a glob pattern to the gcs_list_blobs function


jdddog (approved):
Thanks Keegan, looks good.

PR #634, opened 2023-08-17
Upgrade Terraform and Packer

alexmassen-hane:
PR to upgrade to Terraform v1.5.5 and Packer v1.9.2.

I have done a test deployment to my project and the platform deploys successfully. 
Thankfully all of the terraform scripts work without needing to be changed. I have also updated the documentation to suit the updated package versions. We will need to go onto Terraform Cloud and update the remote version of Terraform to match.


jdddog (approved):
Thanks for upgrading Terraform and Packer Alex, the changes look good.

PR #635, opened 2023-08-21
Added timestamp to json parse

keegansmith21:
Updated the load_and_parse_json() function to parse timestamps separately to dates. Also added custom parsing fields.

Also added some tests for the function.


jdddog (commented):
Hey Keegan, it looks good to me, you could change the `set([])` statements to set literals though.

jdddog (approved):

PR #636, opened 2023-08-29
Fix/Update VM Template File

alexmassen-hane:
This PR updates observatory-image.json VM template file to a newer pck.hcl format. 
This removes a warning when building the observatory instance when deploying the platform.


jdddog (approved):
Thanks Alex, it looks good.

PR #637, opened 2023-09-01
Feature/Add functionality to bq_load_from_memory function

alexmassen-hane:
This PR is to add more input options to this function similar to `bq_load_table`. This will be used in future workflows and they need to be able to select the different types of write dispositions. 
The only workflow that uses this function is the doi workflow. I have made it so the default values for the new inputs won't change anything and should run the same as before.

It was also missing a unit test so I have added one based off of `test_bq_load_table`.


jdddog (approved):
Thanks Alex, it looks good to me.

PR #639, opened 2023-09-12
Feature/Create buckets with roles for unit tests

alexmassen-hane:
For some workflows, usage of Google storage buckets require explicit roles and permissions. For example, OpenAlex requires the following permissions for transferring data between AWS and GCS:
- roles/storage.objectViewer
- roles/storage.legacyBucketWriter

These roles are now needed to be explicitly added to buckets, otherwise unit tests will fail due to permission issues. This PR adds the ability to create Google storage buckets with specific roles for usage in unit tests. 

Buckets are now added to the ObservatoryEnvironment class as a data dictionary where the keys are the bucket_ids and the values are a set of roles required. I have added the roles in such a way that it should not interfere with existing unit tests for the academic observatory workflows.

### DIFF
--- a/observatory-platform/observatory/platform/observatory_environment.py
+++ b/observatory-platform/observatory/platform/observatory_environment.py
@@ -73,7 +73,7 @@ from dataclasses import dataclass
 from datetime import datetime, timedelta
 from http.server import SimpleHTTPRequestHandler, ThreadingHTTPServer
 from multiprocessing import Process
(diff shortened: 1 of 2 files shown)


keegansmith21 (approved):
Thanks Alex, this looks good!

jdddog (approved):
Thanks Alex, looks good.

PR #641, opened 2023-09-22
Fix/Install Packer plugins when building the observatory image

alexmassen-hane:
Forgot to add this with the recent Packer upgrade. The new observatory-image.json.pkr.hcl file relies on external plugins to be installed first before the observatory-image can be built. I did this manually before when I upgraded so I missed it adding it to the platform. 

I've tested it with a fresh Packer installation and I've added a unit test to suit.

Please review and edit as needed.

### DIFF
--- a/observatory-platform/observatory/platform/terraform/terraform_builder.py
+++ b/observatory-platform/observatory/platform/terraform/terraform_builder.py
@@ -175,7 +175,7 @@ class TerraformBuilder:
     def install_packer_plugins(self) -> Tuple[str, str, int]:
         """Install the necessary plugins for packer."""
 
-        # Install the Packer plugins by doing an init on the observatory image config
+        # Install the Packer plugins by doing an init on the observatory image config file
         args = ["packer", "init", "observatory-image.json.pkr.hcl"]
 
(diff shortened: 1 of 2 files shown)


observatory-platform/observatory/platform/terraform/terraform_builder.py
```diff
@@ -172,6 +172,27 @@ def build_terraform(self):
         self.make_files()
         self.platform_runner.make_files()
 
+    def install_packer_plugins(self) -> Tuple[str, str, int]:
+        """Install the necessary plugins for packer."""
+
+        # Install the Packer plugins by doing an init on the observatory image config file
+        args = ["packer", "init", "observatory-image.json.pkr.hcl"]
```
keegansmith21, line 179: Is the image config file always in the same directory that the packer script is being run? Would it be better to supply the absolute path?

jdddog: It is OK because the current working directory is set below `cwd=self.terraform_build_path`

jdddog (approved):
Thanks for fixing this Alex

PR #642, opened 2023-09-27
Add functions required to make use of Airflow TaskGroups and test them

jdddog:
Add functions required to make use of Airflow TaskGroups and test them


keegansmith21 (approved):
Thanks Jamie, these changes look good. I'm looking forward to changing the dag construction of the telescopes!

PR #643, opened 2023-10-16
Added readthedocs config file

keegansmith21:
Added reathedocs config file - which is now required for rtd builds


kathrynnapier (approved):
Thanks @keegansmith21, all looks great to me!

PR #647, opened 2023-11-05
Moving logging for compare_lists_of_dicts into its own function

jdddog:
Separating out this function so that it can be used elsewhere


keegansmith21 (approved):
Thanks for making this change, Jamie :+1:

PR #649, opened 2023-12-15
Temporary fix for incorrect MYSQL GPG Key

jdddog:
Temporary fix for incorrect MySQL GPG Key: https://github.com/apache/airflow/issues/36231

Since we do not use MySQL we can just remove it from sources.


alexmassen-hane (approved):
Looks good! Glad the fix worked.

PR #650, opened 2024-01-09
Airflow MySQL key updates

keegansmith21:
Temporary fix no longer required as per https://github.com/apache/airflow/issues/36231#issuecomment-1858702014


keegansmith21:
@kathrynnapier this is just reverting a change that Jamie made not long ago. There was some issue with GPG keys that has since been fixed.

kathrynnapier (approved):
Thanks Keegan!

PR #654, opened 2024-02-28
Add Astronomer Starship Airflow plugin to enable migration of DAGs to Astro

jdddog:
Setup the Astronomer Starship Airflow plugin to enable migration of DAGs to Astro


keegansmith21 (approved):
Thanks Jamie!

PR #655, opened 2024-03-22
Set Celery version

jdddog:
Set the Celery Airflow providers package version as the newer version doesn't work with Airflow 2.7.3.


keegansmith21 (approved):
Thanks Jamie. Changes look good!

PR #657, opened 2024-05-27
Added function to normalise date strings

keegansmith21:
Added a function to give a generic datetime string for any valid input datetime string. 

I made this because I was running into issues when running tests locally. When uploading to Bigquery, the string will change format to +0000UTC, regardless of what the timezone of the local string is.

The idea is to put both the local and the uploaded version of the strings through the normalising function and they should always be the same if they represent identical datetimes.

### DIFF
--- a/observatory_platform/date_utils.py
+++ b/observatory_platform/date_utils.py
@@ -14,19 +14,22 @@
 
 # Author: Keegan Smith
 
+from datetime import datetime
+from typing import Union
 from zoneinfo import ZoneInfo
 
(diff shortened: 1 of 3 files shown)


jdddog (commented):
Thanks for the PR Keegan.

Perhaps it would be better to use pendulum, which is already installed, rather than installing a new package?

You can use `pendulum.parse("your-date-time", strict=False)` to parse the date.

jdddog (approved):
Thanks Keegan!

PR #658, opened 2024-05-29
Updated naming convention

keegansmith21:
Implemented the following changes:
- Renamed DatasetAPI parameters with bq_ prefixes where relevant
- Renamed dataset_id in DatasetRelease class to `entity_id` and updated schema accordingly

Note that this will require table reuploads for the current oaebu workflow projects due to the schema changes once merged.


jdddog (approved):
Thanks Keegan, these changes look good to me.

PR #659, opened 2024-06-05
Add pendulum datetime handler to date normalisation

keegansmith21:
I foolishly expected the pendulum datetimes and python datetimes to behave the same, but they don't - perhaps because of versioning issues. So I made an extra step to convert the pendulum datetime to a string first.


jdddog (approved):
Thanks Keegan!

PR #660, opened 2024-07-17
Update start date deduction for dag run

keegansmith21:
Use the airflow functions to get the next dagrun interval in the sandbox environment. I did this because custom timetables will not allow you to pass only an execution date to `infer_automated_data_interval`, which is what ends up being called when `create_dagrun` is invoked without a `data_interval`. Passing an execution date will be deprecated in Airflow 3 so we would have had to make this change sooner or later anyway.

Date calculations change because of this so this will probably require a bit of test updating

### DIFF
--- a/observatory_platform/sandbox/sandbox_environment.py
+++ b/observatory_platform/sandbox/sandbox_environment.py
@@ -1,5 +1,3 @@
-# Licensed to the Apache Software Foundation (ASF) under one
-# or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
 # regarding copyright ownership.  The ASF licenses this file
 # to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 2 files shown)


jdddog (commented):
Hey Keegan, thanks for making this change, it looks good.

Since Airflow changed `execution_date` to `logical_date` (for the most part, I still see there are some Airflow functions using `execution_date` , e.g. create_dagrun), I wonder if it is worth renaming `execution_date` to `logical_date` throughout our code base at the same time?

https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dag-run.html#data-interval

image

keegansmith21:
Thanks @jdddog. Yeah I think that's a great Idea. I'll do that

jdddog (approved):

PR #661, opened 2024-07-24
Execution date -> Logical date

keegansmith21:
Updated references to execution_date to logical_date as it's being deprecated. This will definitely break code and tests in oaebu and ao workflows.

Also altered some of the xcom query code to use the timestamp over the execution_date.


jdddog (approved):
Thanks Keegan, the changes look good to me.

PR #663, opened 2024-08-15
Make PreviousDagRunSensor more robust

jdddog:
Fix an issue where PreviousDagRunSensor was not waiting for previous DAG runs because execution_delta was specified in each PreviousDagRunSensor instance and could get out of sync with the schedule interval defined in the DAG. Instead I've defined a custom execution_date_fn that finds all logic dates for DAG runs before the current DAG run.


keegansmith21 (approved):
Thanks Jamie. Looks good!

PR #665, opened 2025-03-19
2.10.5 Airflow bump

keegansmith21:
Upgrade airflow version to 2.10.5

### DIFF
--- a/observatory_platform/dataset_api.py
+++ b/observatory_platform/dataset_api.py
@@ -25,6 +25,7 @@ import pendulum
 from google.cloud import bigquery
 
 from observatory_platform.config import module_file_path
+from observatory_platform.date_utils import datetime_normalise
 from observatory_platform.google.bigquery import (
     bq_create_dataset,
     bq_create_empty_table,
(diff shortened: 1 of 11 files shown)


observatory_platform/dataset_api.py
```diff
@@ -116,14 +117,14 @@ def from_dict(_dict: Dict) -> DatasetRelease:
...
-            changefile_start_date=bq_timestamp_to_pendulum(_dict.get("changefile_start_date")),
-            changefile_end_date=bq_timestamp_to_pendulum(_dict.get("changefile_end_date")),
+            created=_dt(_dict["created"]),
+            modified=_dt(_dict["modified"]),
+            data_interval_start=_dt(_dict.get("data_interval_start")),
+            data_interval_end=_dt(_dict.get("data_interval_end")),
+            snapshot_date=_dt(_dict.get("snapshot_date")),
+            partition_date=_dt(_dict.get("partition_date")),
+            changefile_start_date=_dt(_dict.get("changefile_start_date")),
+            changefile_end_date=_dt(_dict.get("changefile_end_date")),
```
jdddog, line 127: These changes should be reverted, as the BigQuery string timestamps need to be converted to pendulum DateTime objects inside from_dict.

observatory_platform/dataset_api.py
```diff
@@ -139,14 +140,14 @@ def to_dict(self) -> Dict:
...
-            changefile_start_date=pendulum_to_bq_timestamp(self.changefile_start_date),
-            changefile_end_date=pendulum_to_bq_timestamp(self.changefile_end_date),
+            created=_dt(self.created),
+            modified=_dt(self.modified),
+            data_interval_start=_dt(self.data_interval_start),
+            data_interval_end=_dt(self.data_interval_end),
+            snapshot_date=_dt(self.snapshot_date),
+            partition_date=_dt(self.partition_date),
+            changefile_start_date=_dt(self.changefile_start_date),
+            changefile_end_date=_dt(self.changefile_end_date),
```
jdddog, line 150: I think these changes can also be reverted, as they just needed to be converted into iso8601 strings.

observatory_platform/tests/test_dataset_api.py
```diff
@@ -61,7 +61,7 @@ def test_add_dataset_release(self):
             rows = bq_run_query(f"SELECT * FROM `{api.full_table_id}`")
             self.assertEqual(1, len(rows))
             actual = DatasetRelease.from_dict(dict(rows[0]))
-            self.assertEqual(expected, actual)
+            self.assertEqual(expected.to_dict(), actual.to_dict())
```
jdddog, line 64: The actual object should be compared (it is compared with the "__eq__" function), as the dict form is different.

observatory_platform/tests/test_dataset_api.py
```diff
@@ -89,7 +89,7 @@ def test_get_dataset_releases(self):
 
             # Get releases
             actual = api.get_dataset_releases(dag_id=dag_id, entity_id=entity_id)
-            self.assertListEqual(expected, actual)
+            self.assertListEqual([e.to_dict() for e in expected], [a.to_dict() for a in actual])
```
jdddog, line 92: The actual object should be compared (it is compared with the "__eq__" function), as the dict form is different.

observatory_platform/tests/test_dataset_api.py
```diff
@@ -156,26 +156,26 @@ def test_get_latest_dataset_release(self):
                 api.add_dataset_release(release)
 
             latest = api.get_latest_dataset_release(dag_id=dag_id, entity_id=entity_id, date_key="snapshot_date")
-            self.assertEqual(releases[-1], latest)
+            self.assertEqual(releases[-1].to_dict(), latest.to_dict())
```
jdddog, line 159: The actual object should be compared (it is compared with the "__eq__" function), as the dict form is different.

jdddog (changes requested):
Thanks Keegan, that looks good, just a couple of suggested changes related to the DatasetRelease class.

